PR #1, opened 2020-03-02
Initial HTTP API Draft

dvorak42:
This is a basic draft for the HTTP API, it still needs work to bring it into sync with the architecture doc (and to fill out some of the registry/storage sections).

Will keep on iterating on it, but wanted to make the PR so people can start looking at it/responding to it, and also to figure out which bits should be in this doc versus the other PrivacyPass docs.

### DIFF
--- a/drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
+++ b/drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
@@ -1,0 +1,226 @@
+---
+title: "Privacy Pass: HTTP API"
+abbrev: PP http api
+docname: draft-pp-http-api-latest
+date:
+category: info
+
+ipr: trust200902
+keyword: Internet-Draft
+
+stand_alone: yes
+pi: [toc, sortrefs, symrefs]
+
+author:
+ -
+    ins: S. Valdez
+    name: Steven Valdez
+    org: Google LLC
+    email: svaldez@chromium.org
+
+
+informative:
+  RFC5785:
+  RFC4648:
+  draft-ietf-httpbis-header-structure-15:
+  
+--- abstract
+
+This document specifies an integration for Privacy Pass over an HTTP API,
+along with recommendations on how key commitments are stored and accessed
+by HTTP-based consumers.
+
+--- middle
+
+# Introduction
+
+The Privacy Pass protocol as described in {{draft-davidson-pp-protocol}}
+can be integrated with a number of different settings, from server to
+server communication to browsing the internet.
+
+In this document, we will provide an API to use for integrating Privacy
+Pass with an HTTP framework. Providing the format of HTTP requests and
+responses needed to implement the Privacy Pass protocol.
+
+### Terminology
+
+The following terms are used throughout this document.
+
+- Server: A service that provides access to a certain resource
+  (typically denoted S)
+- Client: An entity that seeks authorization from a server (typically
+  denoted C)
+- Key: Server VOPRF key
+- Commitment: Alternative name for Server's public key.
+
+### Protocol messages
+
+Protocol messages are described in the TLS presentation language
+[RFC8446] and will be encoded as raw bytes strings within the containing
+headers.
+
+## Layout
+
+- {{wrapping}}: Describes the wrapping of messages within HTTP
+  requests/responses.
+- {{config-retrieval}}: Describes how HTTP clients retrieve server
+  configurations.
+- {{commitment}}: Describes how HTTP clients retrieve key commitments as
+  part of the issuance/redemption actions.
+- {{issuance}}: Describes how issuance requests are performed via a HTTP
+  API.
+- {{redemption}}: Describes how redemption requests are performed via a
+  HTTP API.
+- {{storage}}: Describes how HTTP clients should store and discard
+  received Privacy Pass tokens.
+
+## Requirements
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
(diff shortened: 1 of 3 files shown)


drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
```diff
@@ -0,0 +1,423 @@
...
+
+### Terminology
+
+The following terms are used throughout this document.
+
+- Server: A service that provides access to a certain resource
+  (typically denoted S)
+- Client: An entity that seeks authorization from a server (typically
+  denoted C)
+- Key: Server VOPRF key
```
alxdavids: This is probably hangover from draft-privacy-pass, but we should remove mention of VOPRFs to be as generic as possible

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
```diff
@@ -0,0 +1,423 @@
...
+3. The client then makes a POST request to
+   <``server_origin``>/.well-known/privacy-pass with a message of type
+   ``request-issuance`` and a body of:
+
+~~~
+enum { Normal(0) } IssuanceType;
+
+struct {
+    IssuanceType type = 0;
+    opaque issue_data<0..2^16-1> = client_issue.issue_data;
```
alxdavids: I guess that this should be `i_data`?

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
```diff
@@ -0,0 +1,423 @@
...
+~~~
+struct {
+    IssuanceType type = request.type;
+    uint16 ciphersuite = server_issue_resp.ciphersuite;
+    opaque evals<1..2^16-1> = server_issue_resp.evals;
+    opaque proof<1..2^16-1> = server_issue_resp.proof;
+}
+~~~
+
+6. The client should then should call the ``CLIENT_ISSUE_FINISH``
```
alxdavids: This is slightly different to how the SERVER_ISSUE interface currently operates, as it this interface that calls CLIENT_ISSUE_FINISH. Perhaps the interface should be modified so that SERVER_ISSUE returns the response to CLIENT_ISSUE_GEN to match what you have here?

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
```diff
@@ -0,0 +1,423 @@
...
+described below.
+
+## Generic Token Redemption {#token-redemption}
+
+Inputs:
+- ``server_id``: The server ID to redeem a token against.
+- ``ciphersuite``: The ciphersuite for this token.
+- ``public_key``: The public key associated with this token.
+- ``token``: A Privacy Pass token.
+- ``additional_data``: Additional data to bind to this token redemption.
```
alxdavids: Eventually we may want to be more prescriptive about how additional_data is formed as it is the HMAC input. This document may be the best place to do that.

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
dvorak42: Yeah, was going to comment on the architecture doc, in that for some of the interfaces its a bit odd that they internally call back and forth between the server and client, which makes it hard to integrate at the communication between the two.

One potential change is making them return these intermediate responses which are then used as inputs to the matching side, and optionally having interfaces that encapsulate the whole flow, so that APIs built on top of them can integrate between the individual steps rather than needing to redefine how they internally call back and forth.

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
dvorak42: Ah yeah, synced with a previous copy of the protocol doc, updated.

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
dvorak42: Specific construction for the delegated case is down below, but I'm not sure whether the client (UA) which originally has the tokens should be doing PP_Redeem and then passing the result of that to whatever website wants to redeem a token (allowing the UA to set the aux additional_data, so that it can't be reused elsewhere), or whether the website should be taking the raw token and doing its own PP_Redeem with whatever it wants to insert in aux.

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
alxdavids: Yeah rewriting the interfaces to return those values synchronously to the calling interface makes sense. I'll write up the change.

drafts/draft-svaldez-pp-http-api/draft-svaldez-pp-http-api.md
```diff
@@ -121,7 +122,53 @@ anonymization guarantees.
...
+    KeyCommitment commitments<1..2^16-1>;
+}
+~~~
+
+The addition to the log is made via a signed message to the log
+operator, which verifies the authenticity against a public key
+associated with that server origin (either via the Web PKI or a
+out-of-band key).
+
+The server SHOULD then store an inclusion proof of the current key
```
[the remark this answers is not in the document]
dvorak42: Yeah, wasn't sure if using a system that supports an inclusion proof should be part of the main architecture doc or for specific instantiations of Privacy Pass. But yeah, you'd get a new inclusion proof when you do the commitment (probably some update to the GLOBAL_CONFIG_UPDATE interface to support this).

dvorak42:
Yeah, I think there's now enough meet on this to merge it as is, and I can make additional PRs to update it/clean it up.

PR #12, opened 2020-04-21
Updated charter text

alxdavids:
Changes account for feedback received during (and following) virtual BoF session, including:

- Scoping protocol to application-layer.
- Considering the case of transferable tokens only.
- Specific points to consider potential consolidation/trust issues.
- Not committing to a specific set of documents.
- Ensuring that Clients can verify Issuer key material.


alxdavids:
Going to merge this now so that we have a working up-to-date version. Further changes can be made in a new PR.

PR #19, opened 2020-05-05
Rename some fields and fix some nits

chris-wood:
Rename IssuanceMessage and RedemptionMessage to IssuanceRequest and RedemptionRequest, respectively. Rename SecretKey to PrivateKey. Change `max_evals` to an integer value, rather than opaque vector.

I also removed company-specific information, which is not really relevant for the document. (That's useful information, but not appropriate for a standards document, I think.)

cc @alxdavids @dvorak42

### DIFF
--- a/draft-davidson-pp-protocol.md
+++ b/draft-davidson-pp-protocol.md
@@ -157,7 +157,7 @@ already in prime-order groups, and constructions are currently being
 drafted in separate standardization processes {{I-D.irtf-cfrg-voprf}}.
 
 The Privacy Pass protocol is split into three stages. The first stage,
-"initialisation," produces the global server configuration that is
+"initialisation", produces the global server configuration that is
 broadcast to (and stored by) all clients. The "issuance" phase provides
 the client with unlinkable tokens that can be used to initiate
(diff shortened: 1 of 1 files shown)


draft-davidson-pp-protocol.md
```diff
@@ -136,41 +136,33 @@ Such cases may only require a lightweight form of challenge (such as
...
+
+The Privacy Pass protocol can be instantiated using a cryptographic 
+primitive known as a verifiable oblivious pseudorandom function (VOPRF) 
+as the authorization mechanism. Such VOPRF protocols can be implemented 
+already in prime-order groups, and constructions are currently being 
+drafted in separate standardization processes {{I-D.irtf-cfrg-voprf}}.
 
 The Privacy Pass protocol is split into three stages. The first stage,
-initialisation, produces the global server configuration that is
+"initialisation," produces the global server configuration that is
```
alxdavids, line 160:
```suggestion
"initialisation", produces the global server configuration that is
```

alxdavids: also, existing typo on my part, but maybe it should be initialization?

PR #27, opened 2020-05-19
Updates based on 17 May comments

alxdavids:


jsalowey:
I think we should merge this change.

alxdavids:
Merging the change following approvals from the chairs on list.

PR #28, opened 2020-05-20
Another take at unlinkability.

chris-wood:
And fix some nits.

### DIFF
--- a/CHARTER.md
+++ b/CHARTER.md
@@ -5,8 +5,8 @@ mechanism for anonymous token creation and redemption. Servers (Issuers)
 create and later verify tokens that are redeemed by an ecosystem of
 clients, such that:
 
-- Any token granted by a given Issuer is indistinguishable from all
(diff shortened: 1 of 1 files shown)


CHARTER.md
```diff
@@ -5,8 +5,8 @@ mechanism for anonymous token creation and redemption. Servers (Issuers)
 create and later verify tokens that are redeemed by an ecosystem of
 clients, such that:
 
-- Any token granted by a given Issuer is indistinguishable from all
-  other tokens granted by the same Issuer key at redemption time.
+- An Issues cannot link a redeemed token to one of N previously granted tokens
```
dvorak42, line 8: *Issuer

chris-wood:
> Hmm, I think this still needs to say something about keys.

Yeah, good point!

alxdavids:
I think this clarification is more specific than the previous wording so I'm happy to add it.

PR #29, opened 2020-06-15
Synchronize charter with datatracker 00-00

kaduk:
Before entering the draft charter into the datatracker for IESG+IAB evaluation (https://datatracker.ietf.org/doc/charter-ietf-privacypass/), I made a few edits.  Let's get them in place here as well, to avoid skew in what people see if they look in different places.

I also put a bit of commentary on the motivation for the changes, at https://github.com/kaduk/privacy-pass-ietf/pull/1/files


dvorak42:
+1, looks good.

PR #41, opened 2020-07-10
Remove unnecessary key rotation section

alxdavids:
cc @dvorak42 @chris-wood


alxdavids:
> LGTM, 1 week might still be too small of a window for actively malicious servers, but I think that ties in to the open issue regarding malicious servers rather than the generic key rotation considerations.

Yeah, I agree. Something to bring up for discussion.

PR #56, opened 2021-01-05
Rename drafts for adoption into WG

alxdavids:
- Also update CONTRIBUTING.md with correct links


dvorak42 (approved):
Looks reasonable. Once this lands we should see about hooking about the gh-pages generation and link to that.

Issue #49, opened 2020-11-25
Token expiration


chris-wood:
There's currently no text around token expiration. Minimally, we could say that tokens expire when issuer (and verifier) keys rotate. Do we also want to bake in mechanisms for expiration to be cryptographically bound to issued tokens? For example, one might imagine using PMB-like constructions to encode an expiration "epoch" into the metadata bits. The single metadata bit variant would allow two epochs that way. 

@dvorak42, @alxdavids: what do you think?

dvorak42:
I'm not sure what baking the expiry into the tokens gives that can't be managed by rotating the issuer/verifier keys (or using public metadata (which is effectively the key used to sign))? That way issuers that care about more fine grained token expiration use the key rotation for that and issuers that only care about coarse token expiration use it for other metadata, allowing the protocol to treat both in the same way?

chris-wood:
Yeah, I think the only two options are (a) rotate keys or (b) include the expiry information as public metadata (and need not be tied to the key used to sign). I'm not advocating for one or the other. 

> That way issuers that care about more fine grained token expiration use the key rotation for that and issuers that only care about coarse token expiration use it for other metadata, allowing the protocol to treat both in the same way?

Sure, that's reasonable. We should probably have text on this point. (That's partly why I filed this issue.)

chris-wood:
We should also make sure to note that the validity window does narrow the anonymity set.

Issue #48, opened 2020-10-16
Describe privacy implications of underlying transport protocol(s)


chris-wood:
Token issuance and redemption flows from the same IP address, for example, would be linkable, but this isn't a break in the privacy pass protocol. We should discuss these orthogonal privacy considerations.

chris-wood:
I think this is already addressed [in the draft](https://github.com/ietf-wg-privacypass/base-drafts/blob/master/draft-ietf-privacypass-architecture.md#client-identifying-information-client-ip). @alxdavids, if you agree, can we close this?

alxdavids:
Yep, good point, closing.

Issue #47, opened 2020-10-16
Describe negative consequences of centralization


chris-wood:
We don't have a good way to address this (now), so we should minimally acknowledge the risks and consequences.

chris-wood:
Dupe of #45. Closing.

PR #60, opened 2021-02-18
Add metadata discussion.

### LINKED ISSUE #51: Expand metadata discussion
There are several dimensions to metadata, including:

- public vs private
- verifiable vs non-verifiable
- bounded or arbitrary-length

It's worth expanding on these dimensions. For example, the underlying cryptographic primitive may influence or limit one or more of these dimensions in a variety of ways. Also, whether metadata is public or private, or verifiable or not, is important for privacy reasons. (If metadata can't be verified, is it a tracking vector? Can private metadata track clients?)

In general, expounding upon these issues would likely be quite valuable for the draft.

chris-wood:
Closes #51, to some extent.

cc @alxdavids

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -457,6 +457,8 @@ is described does not include additional metadata. However, future instantiation
 may use this functionality to provide redemption verifiers with additional
 information about the user.
 
+## Client privacy implications
+
 Note that any metadata bits of information can be used to further
 segment the size of the user's anonymity set. Any server that wanted to
 track a single user could add a single metadata bit to user tokens. For
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -28,18 +28,7 @@ author:
 normative:
   RFC2119:
   RFC8446:
-  draft-davidson-pp-protocol:
-    title: "Privacy Pass: The Protocol"
-    target: https://tools.ietf.org/html/draft-davidson-pp-protocol-00
-    author:
-      ins: A. Davidson
-      org: Cloudflare Portugal
-  draft-svaldez-pp-http-api:
-    title: "Privacy Pass: HTTP API"
-    target: https://github.com/alxdavids/privacy-pass-ietf/tree/master/drafts/draft-svaldez-pp-http-api
-    author:
-      ins: S. Valdez
-      org: Google LLC
+
 informative:
   TrustTokenAPI:
     title: Getting started with Trust Tokens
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -420,12 +424,55 @@ Therefore, we are careful to ensure that the number of servers is kept
...
+generates itself, or the server provides during the issuance flow and
+the client can check for correctness. Opaque public metadata is metadata
+the client can see but cannot check for correctness. As an example, the
+opaque public metadata might be a "fraud detection signal", computed on
+behalf of the server, during token issuance. In normal circumstances,
+clients cannot determine if this value is correct or otherwise a tracking
+vector.
+
+Private metadata is that which clients cannot observe as part of the token
+issuance flow. In {{draft-davidson-pp-protocol}}, it is possible to include
```
alxdavids, line 454: We probably need to update the refs to use the new WG versions.
```suggestion
issuance flow. In {{draft-ietf-privacypass-protocol}}, it is possible to include
```

draft-ietf-privacypass-architecture.md
```diff
@@ -420,12 +424,55 @@ Therefore, we are careful to ensure that the number of servers is kept
...
+private metadata to redemption tokens. The core protocol instantiation that
+is described does not include additional metadata. However, future instantiations
+may use this functionality to provide redemption verifiers with additional
+information about the user.
+
```
alxdavids, line 459: It might be worth separating the last two paragraphs as they relate mostly to privacy considerations. What do you think?

```suggestion

## Client privacy implications

```

draft-ietf-privacypass-architecture.md
```diff
@@ -420,12 +424,55 @@ Therefore, we are careful to ensure that the number of servers is kept
...
+Public metadata is that which clients can observe as part of the token
+issuance flow. Public metadata can either be transparent or opaque. For
+example, transparent public metadata is a value that the client either
+generates itself, or the server provides during the issuance flow and
+the client can check for correctness. Opaque public metadata is metadata
+the client can see but cannot check for correctness. As an example, the
+opaque public metadata might be a "fraud detection signal", computed on
+behalf of the server, during token issuance. In normal circumstances,
+clients cannot determine if this value is correct or otherwise a tracking
+vector.
```
alxdavids, line 451: We may want to provide references to constructions that permit such metadata to be added (links to KLOS20 and the FB paper). However, it might just be better to wait for the CFRG drafts?

alxdavids (commented):
Thanks Chris, this looks great (and thanks for joining me! :D). I have a few comments to address. Also, I hadn't considered the differences between opaque and transparent public metadata, but there is an important distinction there so thanks for bringing it up.

draft-ietf-privacypass-architecture.md
chris-wood: Yeah, good call :-)

draft-ietf-privacypass-architecture.md
chris-wood: Nah, citing them here is a good idea. Let me add them!

chris-wood: Done!

chris-wood:
@alxdavids updated with citations. I'll take #49, #48, and #47 next. (Trying to get these done before the upcoming deadline.)

alxdavids (approved):
LGTM! I'll take some of the other issues.

PR #61, opened 2021-02-18
Add token expiration discussion.

### LINKED ISSUE #49: Token expiration

chris-wood:
Replace suggested registry with list of other options. The rationale is that a registry may be sufficient but not necessary. There are other options that might work. 

I suggest we work with @chelseakomlo on a draft describing these techniques in more detail.

cc @alxdavids 

Closes #49.

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -249,7 +249,7 @@ SHOULD minimize or remove identifying information where possible, e.g.,
 by using anonymity-preserving tools such as Tor to interact with
 servers.
 
-# Key management framework {#key-mgmt}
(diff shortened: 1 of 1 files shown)


alxdavids (approved):
This LGTM. Agreed that working on a document that tackles these questions specifically would be worthwhile.

Issue #50, opened 2020-12-02
Accommodating two round protocols


chris-wood:
Some protocols, like [blind Schnorr signature schemes](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.3.6950&rep=rep1&type=pdf), require more than 1 round trip to complete. Currently, the protocol assumes a protocol that runs in one round trip, with some offline public key distribution step. In particular, issuance request messages are mapped almost directly to the underlying OPRF blind/request message. 

We should consider what needs to change, if anything, to accommodate these differently shaped protocols. I suggest we loosen the 1-1 restriction (?) from the Privacy Pass API to the underlying cryptographic protocol APIs. It just so happens that Privacy Pass maps nicely to the OPRF construction, but that need not be the case. For example, and hand waving over details, in the blind Schnorr case, an implementation of the issuance request flow could involve first fetching the nonce commitment from the server, and then sending a blinded value for evaluation. 

@dvorak42, what do you think?

chris-wood:
@alxdavids what do you want to do here? Should we refactor the protocol to run in two rounds? Perhaps something like the following:

- CommitRequest
- CommitResponse
- IssueRequest
- IssueResponse

During the commitment flow (first round trip), the server can provide any sort of nonce or whatever is needed for the underlying protocol.

alxdavids:
I think that sounds fine. Presumably in this architecture, the CommitRequest from the Client will mostly be an empty message (at least for 3-round protocols)?

dvorak42:
For 1 round protocols, would the key commitment/configuration could be used to provide a 'default' result to the CommitResponse, either hardcoded as part of the version (or some parameter in the configuration)?

chris-wood:
> For 1 round protocols, would the key commitment/configuration could be used to provide a 'default' result to the CommitResponse, either hardcoded as part of the version (or some parameter in the configuration)?

Not sure! We'd have to play with it and see what works.

chris-wood:
@alxdavids will draft a PR based on the commit request/response flow, where the response includes some implementation-specific handle to state needed for the subsequent issuance flow. For one round protocols, the handle will be a fixed (empty?) string passed to the issuance flow.

alxdavids:
I think this has been fixed by #62, @chris-wood to confirm?

PR #62, opened 2021-02-19
First attempt at solving #50.

alxdavids:
- Add an extra round to the issuance phase
- Specify whether commitments are required in ciphersuites
- Allow possibility for null state values

cc @claucece, @chris-wood, @dvorak42, @armfazh for review.


draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
   store[server.id].push(tokens)
 ~~~
 
+Note that the first round of the protocol is only necessitated for
+certain ciphersuites that require prior state generation. When such
```
chris-wood, line 303:
```suggestion
certain ciphersuites that require client and servers commit to some state. When such
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+state `st` is generated and sent to the client, the client returns the
+`st` with the `IssuanceRequest` message. The server must check that the
```
chris-wood, line 305:
```suggestion
`st` with the `IssuanceRequest` message. The server MUST check that the
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+state corresponds to the `st` that was previously sent state. This
```
chris-wood, line 306:
```suggestion
state corresponds to `st` that was previously committed. This
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+requires architecture for remaining stateful, or by passing the state as
+an encrypted (and authenticated) blob. Such statefulness must be
+implemented in similar ways to how TLS session resumption is managed
+{{RFC8446}}.
```
chris-wood, line 310:
```suggestion
requires the commitment to either be a reference to some state on the server, 
or the commitment be an encrypted (and authenticated) blob that the server can
use to recover state. The mechanism by which servers handle this state is implementation
specific, and similar to how TLS session resumption state is managed; see {{RFC8446}}
for details.
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+
+When the server does not need to generate this state, the client instead
+DOES NOT send the CommitRequest message, and runs:
+
+~~~
+cInput = Generate(m, null)
```
chris-wood, line 316:
```suggestion
cInput = Generate(m, "")
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+~~~
+
+A server that is expecting some non-null `st` to be passed must abort
```
chris-wood, line 319:
```suggestion
A server that is expecting some non-empty `st` to be passed must abort
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+the protocol on receiving a request containing a null `st` value.
```
chris-wood, line 320:
```suggestion
the protocol on receiving a request containing an empty `st` value.
```

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,29 @@ authorization tokens from the server.
...
+
+Note: currently, no ciphersuites are supported that support working with
+non-null state messages.
```
chris-wood, line 323:
```suggestion
empty state messages.
```

draft-ietf-privacypass-protocol.md
```diff
@@ -380,6 +415,30 @@ opaque PublicKey<1..2^16-1>
...
+  opaque server_id<1..2^16-1>
+} CommitRequest;
+~~~
+
+### CommitResponse {#pp-cli-commit-response}
+
+The `CommitResponse` struct is contains an opaque set of bytes that
+correspond to some state that the server has generated. How this
+sequence of bytes is generated corresponds to the ciphersuite and
+whether the server is stateful or not.
```
chris-wood, line 434:
```suggestion
correspond to some state that the server has generated. The structure
and format of this value is implementation specific depending on whether
the server is stateful.
```

draft-ietf-privacypass-protocol.md
```diff
@@ -814,22 +878,30 @@ corresponds to the maximum difficulty of computing a discrete logarithm
 in the group. Note that the actual security level MAY be lower. See the
 security considerations in {{I-D.irtf-cfrg-voprf}} for examples.
 
+The COMMIT parameter refers to whether the first round of the issuance
+phase of the protocol is necessary. When this set to false, the client
```
chris-wood, line 882:
```suggestion
phase of the protocol is necessary. When this is false, the client
```

draft-ietf-privacypass-protocol.md
```diff
@@ -814,22 +878,30 @@ corresponds to the maximum difficulty of computing a discrete logarithm
...
+ignores the first message and uses a null value for the state parameter
+sent by the server.
```
chris-wood, line 884:
```suggestion
ignores the first message and uses an empty value for the state parameter
to `Generate`.
```

draft-ietf-privacypass-protocol.md
```diff
@@ -380,6 +415,30 @@ opaque PublicKey<1..2^16-1>
...
 ~~~
 
+### CommitRequest {#pp-cli-commit-request}
+
+The `CommitRequest` struct is simply a fixed message corresponding to
+the ID of the server.
+
+~~~
+struct {
+  opaque server_id<1..2^16-1>
```
chris-wood, line 425: Interesting -- do we need this? Can it be empty?

chris-wood (changes requested):
This looks good! I think we can shift the presentation of the `st` value to be something about a reference to some state, rather than actual state itself. That allows the server to use it as a reference (a key in some database) or as state (encrypted cookie).

draft-ietf-privacypass-protocol.md
alxdavids: I was allowing for the possibility that the client may only want to allow commitments for certain types (e.g. it may only want to allow server S with ciphersuite `c1`, rather than `c2`). The `server_id` was kind of a placeholder for that. However, it is pretty ugly right now, so I would be happy to remove it.

alxdavids:
Thanks @chris-wood, changes all look good. What do you think about the `GenerateState` algorithm? Maybe I can just add some text that leaves this up to the implementer?

chris-wood:
That's probably only a function that 3/4-message protocols need to implement, right? I wouldn't expect the VOPRF draft to implement it, since COMMIT=false for those suites.

Also, what do you think about renaming it to GenerateCommitment?

alxdavids:
> That's probably only a function that 3/4-message protocols need to implement, right? I wouldn't expect the VOPRF draft to implement it, since COMMIT=false for those suites.
> 
> Also, what do you think about renaming it to GenerateCommitment?

Sounds good, I've also replaced `state` with `commitment` throughout to avoid mishaps.

draft-ietf-privacypass-protocol.md
chris-wood: 👍 Eh, it's probably fine for now.

draft-ietf-privacypass-protocol.md
```diff
@@ -287,6 +299,33 @@ authorization tokens from the server.
   store[server.id].push(tokens)
 ~~~
 
+Note that the first round of the protocol is only necessitated for
+certain ciphersuites that require client and servers commit to some
+value. When such commitment `com` is generated and sent to the client,
+the client returns `com` with the `IssuanceRequest` message. The server
+MUST check that the commitment corresponds to `com` that was previously
```
claucece, line 306: When the commitment does not corresponds to `com` that was previously
committed, should this define a specific error that we need to add to the `Error types`?

claucece:
I like this ;) Thank you @alxdavids ! But, should the `CommitRequest and `GenerateCommitment` should be added to the `API functions` sections, as `Generate` and others?

draft-ietf-privacypass-protocol.md
alxdavids: @claucece Good point! Yes it should be.

alxdavids:
> 
> 
> I like this ;) Thank you @alxdavids ! But, should the `CommitRequest and `GenerateCommitment`should be added to the`API functions`sections, as`Generate` and others?

Yeah I think you're right. I didn't add them as I wasn't exactly sure how they should be implemented, but I guess that we can leave that for the underlying ciphersuites to decide.

chris-wood:
Yeah, maybe just leave an OPEN ISSUE noting that they should be defined? If you want something, maybe this would work? (I replaced server_id with metadata, though I wonder if client-provided metadata is best supplied during the commit flow or the issuance flow?) 

```
Prepare

A function run by the client to prepare for a commitment will used during the issuance flow of the Privacy Pass protocol.

Inputs:

info: An opaque byte application-specific byte string.

None

Outputs:

commit_req: A CommitRequest struct.



Commit

A function run by the server to commit to a Privacy Pass issuance flow.

Inputs:

pkS: A server PublicKey.
skS: A server PrivateKey.
commit_req: A CommitRequest struct.

Outputs:

commit_resp: A CommitResponse struct.
```

alxdavids:
@chris-wood Thanks for the example, I've altered the PR to include this and @claucece's feedback.

chris-wood:
Looks good! Shall we merge and the submit this and -architecture?

alxdavids:
> 
> 
> Looks good! Shall we merge and the submit this and -architecture?

SGTM! I'm AFK for the rest of today, but I'll handle it tomorrow if that's okay with everyone else?

claucece:
@alxdavids thank you! looking good ;)

Issue #77, opened 2021-07-09
Pin protocol messages to underlying cryptographic scheme


chris-wood:
Protocols like VOPRF and blind RSA/BLS have different semantics and wire images. As an example, batching seems to be a per-crypto-protocol property. VOPRF supports it, whereas BLS/RSA do not.

Trying to map these to a shared wire format seems a bit awkward. I recommend we parameterize all PP protocol messages based on the underlying cryptographic construction. So, for example:

```
enum {
	VOPRF,
	BlindRSA,
} Protocol;

struct {
	Protocol proto;
	uint16 length; // allows skipping over unknown protocols
	select (proto) {
		case VOPRF:
			opaque blind_elements<1..2^16-1>;
  		case BlindRSA:
  			opaque blinded_msg[kLen];
	}
} IssuanceRequest;

struct {
	Protocol proto;
	uint16 length; // allows skipping over unknown protocols
	select (proto) {
		case VOPRF:
			IssuedToken tokens[m];
  			opaque proof<1..2^16-1>;
  		case BlindRSA:
  			opaque blind_sig[kLen];
	}
} IssuanceResponse;
```

(The same should probably apply to redemption messages as well.)

We can then refactor the document to only specify how these protocol messages are constructed and parsed, rather than specify how implementation-specific details should work.

chris-wood:
Closing as overcome by events.

PR #79, opened 2021-07-16
Significant refactor to focus on protocol messages and interactions.

chris-wood:
Starting as a draft to demonstrate what I had in mind for #77.

This removes a lot of text, shifting the focus away from client and server implementation-specific APIs towards the protocol messages exchanged during issuance and redemption, and their encoding. This does seem to simplify a lot, as it encapsulates all of the special features of different underlying cryptographic protocols (public verifiability, metadata, etc) to those protocols. The result is just a thin wrapper around these schemes with logic specific to and necessary for privacy pass, i.e., generating nonces as token inputs, publishing server configurations, and so on.

It's not complete, but I wanted to throw it up early to get some preliminary feedback.

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -39,18 +39,6 @@ normative:
   RFC2119:
   RFC8446:
   I-D.irtf-cfrg-voprf:
-  draft-davidson-pp-architecture:
-    title: "Privacy Pass: Architectural Framework"
-    target: https://github.com/alxdavids/privacy-pass-ietf/tree/master/drafts/draft-davidson-pp-architecture
-    author:
-      ins: A. Davidson
-      org: Cloudflare
-  draft-svaldez-pp-http-api:
-    title: "Privacy Pass: HTTP API"
-    target: https://github.com/alxdavids/privacy-pass-ietf/tree/master/drafts/draft-davidson-pp-architecture
-    author:
-      ins: S. Valdez
-      org: Google
 informative:
   RFC7049:
   RFC7159:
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -230,22 +218,48 @@ clients and servers supporting the Privacy Pass protocol.
...
+of this document are written in the TLS presentation language {{RFC8446, Section 3}}.
+
+## Variants {#variants}
+
+This document specifies a single Privacy Pass variant, defined by the
+following enumeration:
+
+~~~
+enum {
+  VOPRF_decaf448_shake256 = 0x0001,
```
dvorak42, line 258: Any chance we can add the p384_sha512 variant?

chris-wood: Yeah, I was just getting one down to start.

draft-ietf-privacypass-protocol.md
```diff
@@ -372,306 +398,220 @@ auxiliary data chosen by the client. The usage of `current_time()`
...
+} VOPRFIssuanceRequest;
 ~~~
 
-### RedemptionResponse {#pp-redemption-response}
+Upon receipt of a VOPRFIssuanceRequest denoted `issuance_req`, the server evaluates it
+to produce a response as follows:
+
+```
+evaluated_elements, proof =
+  context.Evaluate(skS, pkS, issuance_req.blinded_elements)
```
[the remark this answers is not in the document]
chris-wood: Yeah, good point. We should probably length prefix each of the element lists in request and response to make this easier to address later on.

PR #83, opened 2021-09-29
Initial attempt at integration doc

alxdavids:
This is my first draft of an integration document for replacing the current protocol doc. I've tried to keep it as concise as possible. A few questions:

- Would it be better to be more specific about how the POPRF algorithms are used? Or is this level of detail sufficient?
- Can we leave considerations about how public keys are distributed and consumed to the HTTP API doc?
- Is it reasonable to leave metadata-level considerations to the architecture doc.

Let me know what you think so far, cc @chris-wood, @bemasc, @dvorak42, @jsalowey, @claucece, @armfazh

### DIFF
--- a/draft-ietf-privacypass-integration.md
+++ b/draft-ietf-privacypass-integration.md
@@ -81,9 +81,13 @@ before being sent across the wire.
 # Privacy Pass flow {#protocol-flow}
 
 There are three phases in the protocol: the initialization phase, the
-issuance phase, and the redemption phase.
+issuance phase, and the redemption phase. We construct each phase based
+on the POPRF protocol construction detailed in {{I-D.irtf-cfrg-voprf}},
+and all algorithms and data types are inherited as such.
 
-In the initialization phase, the server samples a keypair and publishes
+## Initialization phase
+
+The server samples a keypair and publishes
 a key configuration in a way that clients can retrieve it. This configuration consists
 of the server's public key and configuration information for the underlying POPRF.
 
@@ -95,58 +99,83 @@ struct {
 } KeyConfig;
 ~~~
 
-KeyConfig.suite corresponds to a POPRF ciphersuite from {{I-D.irtf-cfrg-voprf, Section 4}},
-and KeyConfig.public_key corresponds to a serialized public key of length `Ne` bytes 
-(denoted as a `SerializedElement` in {{I-D.irtf-cfrg-voprf, Section 2}}).
+KeyCOnfig.version corresponds to the version of the key configuration
+that is being used, this parameter is left generic and may be altered if
+extensions to the protocol or server configurations are made.
+KeyConfig.suite corresponds to a POPRF ciphersuite from
+{{I-D.irtf-cfrg-voprf, Section 4}}, and KeyConfig.public_key corresponds
+to a serialized public key of length `Ne` bytes (denoted as a
+`SerializedElement` in {{I-D.irtf-cfrg-voprf, Section 2}}). In
 
-In the issuance phase:
+In order for higher-level applications to indicate which key
+configuration is being used, a common identifier, such as 
+`id=SHA256(KeyConfig)`, should be used.
 
-- The client and server optionally agree on some public ``metadata``.
-- The client retrieves an servers public key, and generates some initial
-  ``token`` data. The client cryptographically ``blinds`` this data, and
-  sends it to the server.
-- The server ``signs`` the blinded token data (including optional), and
-  produces a proof that it used the committed keypair, and sends the
-  signature and proof back to the client.
-- The client ``verifies`` the proof, ``unblinds`` the signature, and
-  stores an authenticated triple of the token, optional metadata, and
-  unblinded signature.
+## Issuance phase
 
-In the redemption phase:
+Let `info` be the agreed upon
+metadata between client and server, and let `config` be the server's
+chosen key configuration.
 
-- The client retrieves an authenticated token, optional metadata and
-  signature triple, and sends it to the server.
-- The server ``validates`` that the pair is authenticated correctly and
-  authorizes the client.
+First, a client configures its verifiable context using `config`:
 
-# Partially Oblivious Pseudorandom Function Protocol
+~~~
+client_context = SetupVerifiableClient(config.suite, config.public_key)
+~~~
 
-We can instantiate the protocol flow in {{protocol-flow}} using the
-partially oblivious pseudorandom function (POPRF) protocol in
-{{I-D.irtf-cfrg-voprf}}. In summary, the issuance phase corresponds to
-receiving a pseudorandom function evaluation on the blinded data (with
-optional metadata). The redemption phase corresponds to revealing
-`finalized` data back to the original issuing server.
+Likewise, the server creates its own context using `config` and the
+corresponding private key `key`:
 
-Note that this instantiation only provides a symmetric verification
-mechanism, since the verification of redemptions can only be performed
-by the server possessing the secret issuing key. In
-{{I-D.ietf-privacypass-architecture}}, we provide alternative frameworks
-for allowing asynchronous and delegated verification of tokens.
+~~~
+server_context = SetupVerifiableServer(config.suite, key, config.public_key)
+~~~
 
-## Security guarantees
+The client then creates an issuance request for a random value `nonce` as follows:
 
-The privacy of clients is determined by the unlinkability of client
-requests during the POPRF protocol. Moreover, the one-more-forgery
-security of the POPRF prevents clients from forging valid tokens for a
-given server. See {{I-D.irtf-cfrg-voprf}} for more details.
+~~~
+nonce = random(32)
+blind, blindedElement = context.Blind(nonce)
+~~~
 
-## Metadata
+The client then sends `blindedElement` to the server. The server, upon
+receipt, evaluates the request:
 
-The POPRF protocol provides mechanisms for embedding public metadata
-into the function evaluations. Such metadata should be agreed apriori by
-clients and servers, and is regarded as being public to entities that
-are not even included in the explicit issuance and redemption exchanges.
+~~~
+evaluatedElement, proof = server_context.Evaluate(key, config.public_key, blindedElement, info)
+~~~
+
+The server sends both `evaluatedElement` and `proof` to the client.
+These are concatenated together. As the length of both is fixed, there
+is no ambiguity in parsing the result.
+
+The client then completes issuance as follows:
+
+~~~
+output = client_context.Finalize(nonce, blind, evaluatedElement, info):
+~~~
+
+This procedure may fail with an error (`VerifyError` or
+`DeserializeError`), in which case the issuance is said to have failed.
+The output of the issuance protocol is the concatenation of `nonce` and
+`output`, denoted as `token`:
+
+~~~
+struct {
+   uint8 nonce[32];
+   uint8 output[Nh]; // Nh is as defined in {{I-D.irtf-cfrg-voprf}}
+} Token;
+~~~
+
+## Redemption phase
+
+The client sends the `Token` to the server to
+verify locally. In particular, the server verifies the `Token` as follows:
+
+~~~
+valid = server_context.VerifyFinalize(key, token.nonce, token.output, info)
+~~~
+
+Redemption is considered successful if `valid` is true.
 
 # Protocol ciphersuites {#ciphersuites}
 
--- a/draft-ietf-privacypass-integration.md
+++ b/draft-ietf-privacypass-integration.md
@@ -107,15 +107,11 @@ POPRF.
 
 ~~~
 struct {
-   uint8 version;
    uint16 suite;
    uint8 public_key[Ne];
 } KeyConfig;
 ~~~
 
-KeyConfig.version corresponds to the version of the key configuration
-that is being used, this parameter is left generic and may be altered if
-extensions to the protocol or server configurations are made.
 KeyConfig.suite corresponds to a POPRF ciphersuite from
 {{I-D.irtf-cfrg-voprf, Section 4}}, and KeyConfig.public_key corresponds
 to a serialized public key of length `Ne` bytes (denoted as a
@@ -154,7 +150,7 @@ as follows:
 
 ~~~
 nonce = random(32)
-blind, blindedElement = context.Blind(nonce)
+blind, blindedElement = client_context.Blind(nonce)
 ~~~
 
 The client then sends `blindedElement` to the server. The server, upon
--- a/draft-ietf-privacypass-integration.md
+++ b/draft-ietf-privacypass-integration.md
@@ -1,0 +1,147 @@
+---
+title: "Privacy Pass Protocol Specification"
+abbrev: PP integration
+docname: draft-ietf-privacypass-integration-latest
+date:
+category: info
+
+ipr: trust200902
+keyword: Internet-Draft
+
+stand_alone: yes
+pi: [toc, sortrefs, symrefs]
+
+author:
+ -
+    ins: A. Davidson
+    name: Alex Davidson
+    org: Brave Software
+    city: Lisbon
+    country: Portugal
+    email: alex.davidson92@gmail.com
+
+normative:
+  RFC2119:
+  RFC8446:
+  I-D.irtf-cfrg-voprf:
+  I-D.ietf-privacypass-architecture:
+  I-D.ietf-privacypass-http-api:
+informative:
+  RFC7049:
+  RFC7159:
+
+--- abstract
+
+This document specifies the necessary integration for building the
+Privacy Pass protocol, supporting symmetric verification and public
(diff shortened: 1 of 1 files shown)


bemasc:
I think this doc might be the right place to declare an IANA registry of Privacy Pass ciphersuites, and explain how entries in that registry map to the VOPRF draft, and are represented in the HTTP API.

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+The following terms are used throughout this document.
+
+- Client: An entity that provides authorization tokens to services
+  across the Internet, in return for authorization.
+- Server: A service (also known as an Issuer) that issues Privacy Pass
+  tokens to clients.
+- Key: The secret key used by the server for authorizing client data.
+- Blinding/unblinding: A pair of cryptographic operations that allow
+  randomly distributing and revealing some underlying `data`, whilst
+  preserving the ability to generate valid signatures on `data` itself.
```
chris-wood, line 79:
```suggestion
```

I don't think we need to define this?

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+protocol.
+
+--- middle
+
+# Introduction
+
+The Privacy Pass protocol provides a privacy-preserving authorization
+mechanism. In essence, the protocol allows clients to provide
+cryptographic tokens that prove nothing other than that they have been
+authorized by a given server in the past
```
chris-wood, line 47:
```suggestion
created by a given server in the past
```

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+{{I-D.ietf-privacypass-architecture}}.
+
+This document provides the necessary integration for building the
+authorization framework, based on existing constructions of oblivious
+pseudorandom function protocols {{I-D.irtf-cfrg-voprf}}. Moreover, we
+show how this integration allows `public` metadata to be introduced to
```
chris-wood, line 53:
```suggestion
show how this integration allows public metadata to be introduced to
```

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+We assume that all protocol messages are encoded into raw byte format
+before being sent across the wire.
+
+# Privacy Pass flow {#protocol-flow}
+
+There are three phases in the protocol: the initialization phase, the
+issuance phase, and the redemption phase.
+
+In the initialization phase, the server samples a keypair and publishes
+a ``public key`` (or key commitment) to a public repository.
```
chris-wood, line 90:
```suggestion
a key configuration in a way that clients can retrieve it. This configuration consists
of the server's public key and configuration information for the underlying POPRF.

~~~
struct {
   uint8 version;
   uint16 suite;
   uint8 public_key[Ne];
} KeyConfig;
~~~

KeyConfig.suite corresponds to a POPRF ciphersuite from {{I-D.irtf-cfrg-voprf, Section 4}},
and KeyConfig.public_key corresponds to a serialized public key of length `Ne` bytes 
(denoted as a `SerializedElement` in {{I-D.irtf-cfrg-voprf, Section 2}}).
```

chris-wood: It's hard to interop if we don't describe how configurations are represented, I think.

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+
+In the issuance phase:
```
chris-wood, line 92: I would spell this differently. For example, I might literally frame it in terms of the POPRF protocol, like so:

```
The issuance phase proceeds as follows. Let `info` be the agreed upon metadata between client
and server, and let `config` be the server's chosen key configuration.

First, a client configures its verifiable context using `config`:

~~~
client_context = SetupVerifiableClient(config.suite, config.public_key)
~~~

Likewise, the server creates its own context using `config` and the corresponding private key `key`:

~~~
server_context = SetupVerifiableServer(config.suite, key, config.public_key)
~~~

The client then creates an issuance request as follows:

~~~
nonce = random(32)
blind, blindedElement = context. Blind(nonce)
~~~

The client then sends blindedElement to the server. The server, upon receipt, evaluates the request:

~~~
evaluatedElement, proof = server_context.Evaluate(key, config.public_key, blindedElement, info)
~~~

The server sends both `evaluatedElement` and `proof` to the client. These are concatenated together. As the length of both is fixed, there is no ambiguity in parsing the result.

The client then completes issuance as follows:

~~~
output = client_context.Finalize(nonce, blind, evaluatedElement, info):
~~~

This procedure may fail with an error (`VerifyError` or `DeserializeError`), in which case the issuance is said to have failed. The output of the issuance protocol is the concatenation of `nonce` and `output`, denoted as `token`:

~~~
struct {
   uint8 nonce[32];
   uint8 output[Nh]; // Nh is as defined in {{I-D.irtf-cfrg-voprf}}
} Token;
~~~

In the redemption phase, the client sends the Token to the server to verify locally. In particular, the server verifies the Token as follows:

~~~
valid = server_context.VerifyFinalize(key, token.nonce, token.output, info)
~~~

Redemption is considered successful if `valid` is true.
```

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+  unblinded signature.
+
+In the redemption phase:
+
+- The client retrieves an authenticated token, optional metadata and
+  signature triple, and sends it to the server.
+- The server ``validates`` that the pair is authenticated correctly and
+  authorizes the client.
+
+# Partially Oblivious Pseudorandom Function Protocol
```
chris-wood, line 112: I would remove this section entirely.

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+## Metadata
+
+The POPRF protocol provides mechanisms for embedding public metadata
+into the function evaluations. Such metadata should be agreed apriori by
+clients and servers, and is regarded as being public to entities that
+are not even included in the explicit issuance and redemption exchanges.
+
+# Protocol ciphersuites {#ciphersuites}
+
+Ciphersuite negotiation is only relevant in terms of negotiating the
```
chris-wood, line 143: Negotiation is a bit odd here, since nothing's negotiated. Maybe this should say that ciphersuites are determined by the server's choice of configuration?

draft-ietf-privacypass-integration.md
chris-wood: We should probably also indicate how the server knows which KeyConfig to use for a given request. I don't think this needs to be part of the protocol here, as it may vary. Instead, I'd recommend coming up with a unique way to identify a KeyConfig, e.g., id = SHA256(KeyConfig), and then saying that the wrapper protocol around Privacy Pass can send this to the server if needed for the given application.

chris-wood (changes requested):
This more or less matches what I was thinking, but I think we can spell things a bit more specifically to enable interop.

draft-ietf-privacypass-integration.md
chris-wood: Version of this key config, so it should probably also include a length of the struct? 🤷

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,147 @@
...
+In the issuance phase:
+
+- The client and server optionally agree on some public ``metadata``.
+- The client retrieves an servers public key, and generates some initial
+  ``token`` data. The client cryptographically ``blinds`` this data, and
+  sends it to the server.
+- The server ``signs`` the blinded token data (including optional), and
+  produces a proof that it used the committed keypair, and sends the
+  signature and proof back to the client.
+- The client ``verifies`` the proof, ``unblinds`` the signature, and
```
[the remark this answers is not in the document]
chris-wood: Yeah, the API should probably just expose a single blob as the token, rather than separate parts.

chris-wood:
@alxdavids, can you please review and respond to the comments above?

alxdavids:
Thanks @chris-wood @dvorak42 @bemasc for the review and suggestions. I've updated the draft with your changes (@chris-wood and @dvorak42 I've also added you as authors due to contributing concrete text). What do you think?

@bemasc: Is the IANA considerations section that I've added sufficient?

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,224 @@
...
+valid = server_context.VerifyFinalize(
+         key, token.nonce, token.output, info
+        )
+~~~
+
+Redemption is considered successful if `valid` is true.
+
+# IANA Considerations
+
+This specification registers new ciphersuites for the Privacy Pass
```
bemasc, line 211: In general the form is: "This specification creates a new page named X in the Y category containing a registry named Z, with the following fields:".

In the case, maybe X = Privacy Pass, Y = Privacy Pass, Z = "Cipher Suites".

Also think about whether your cipher suites need a numeric identifier, or whether it is sufficient to give them strings.

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,224 @@
...
+                  config.suite, key, config.public_key
+                 )
+~~~
+
+The client then creates an issuance request for a random value `nonce`
+as follows:
+
+~~~
+nonce = random(32)
+blind, blindedElement = context.Blind(nonce)
```
chris-wood, line 157:
```suggestion
blind, blindedElement = client_context.Blind(nonce)
```

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,224 @@
...
+## Initialization phase
+
+The server samples a keypair and publishes a key configuration in a way
+that clients can retrieve it. This configuration consists of the
+server's public key and configuration information for the underlying
+POPRF.
+
+~~~
+struct {
+   uint8 version;
```
chris-wood, line 110:
```suggestion
```

chris-wood: I guess one can version above this, so maybe it's not necessary here.

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,224 @@
...
+   uint16 suite;
+   uint8 public_key[Ne];
+} KeyConfig;
+~~~
+
+KeyConfig.version corresponds to the version of the key configuration
+that is being used, this parameter is left generic and may be altered if
+extensions to the protocol or server configurations are made.
```
chris-wood, line 118:
```suggestion
```

draft-ietf-privacypass-integration.md
chris-wood: Why do we need these ciphersuites at all? Aren't they redundant with the POPRF suite?

chris-wood:
> Thanks @chris-wood @dvorak42 @bemasc for the review and suggestions. I've updated the draft with your changes (@chris-wood and @dvorak42 I've also added you as authors due to contributing concrete text). What do you think?

The new text looks mostly good to me, modulo a couple more suggestions and a question about the ciphersuites. @alxdavids, why do we need PP suites when we have POPRF suites?

draft-ietf-privacypass-integration.md
bemasc: I think we do need a new IANA registry.  Without a registry, we would have to Update one or more RFCs every time we want to add a ciphersuite, which is laborious to execute and confusing to read.

Where that registry should be created depends on its intended contents.  If all future VOPRF suites as conceived in the VOPRF draft will continue to be exactly what you want for Privacy Pass, then the VOPRF draft can open the registry.  Otherwise (e.g. if we envision extending Privacy Pass to publicly verifiable ciphersuites), we'll need to define it here.

At the moment, the latter seems more likely.

dvorak42: Yeah, for example in a future world where we have a Blind BLS suite and a POPRF suite, it seems like we'll want some containing ciphersuite that refers to one or the other.

draft-ietf-privacypass-integration.md
```diff
@@ -0,0 +1,224 @@
...
+
+This specification registers new ciphersuites for the Privacy Pass
+protocol, corresponding to the following ciphersuites detailed in
+{{I-D.irtf-cfrg-voprf}}.
+
+- Ciphersuite OPRF(ristretto255, SHA-512) = XXXX
+- Ciphersuite OPRF(decaf448, SHAKE-256) = XXXX
+- Ciphersuite OPRF(P-256, SHA-256) = XXXX
+- Ciphersuite OPRF(P-384, SHA-384) = XXXX
+- Ciphersuite OPRF(P-521, SHA-512) = XXXX
```
dvorak42, line 219: Assuming we want to do a separate IANA registry, we might want to pare down the ones that are initially defined?

draft-ietf-privacypass-integration.md
chris-wood: Sorry, I don't really follow. Why would those things be in the same registry? This document is really just a thin wrapper around the POPRF, and no updates would be needed if new POPRF suites were added. In other words, the POPRF document owns the registry for POPRF suites.

bemasc: Perhaps PMBTokens (an alternative to VOPRF that also matches the synchronous model) is a clearer example than Blind Signatures (which can also be made to fit into the synchronous model, but perhaps less elegantly).

chris-wood: I see your point, but it's not clear to me we actually want a single registry to list different "modes" of privacy pass, e.g., one without private metadata (POPRF) and one with private metadata (PMBTokens). Why not just make that a separate thing altogether? What's the value of merging these things?

dvorak42: At some layer we need to indicate what protocol is being used for a web/HTTP API. Be it PrivacyPassVOPRFP521 or PrivacyPassCoolPostQuantum or PrivacyPassBLS. Seems like something that the PrivacyPass WG should define somewhere as things that higher level things point to, though maybe we don't have that in the Integration doc and put it somewhere else?

bemasc: The goal mentioned in the proposed schedule is to support ecosystems containing a mixture of clients and issuers that support POPRF and non-POPRF ciphersuites.  That means one way or another, different ciphersuite families need to be accessible from the same server identity.

For clarity, it might be more helpful to frame this question in terms of the specific structs in the HTTP API draft.

chris-wood: > The goal mentioned in the proposed schedule is to support ecosystems containing a mixture of clients and issuers that support POPRF and non-POPRF ciphersuites. That means one way or another, different ciphersuite families need to be accessible from the same server identity.

Sure, but they could so by using separate types of key configurations or otherwise making this decision at the application layer. The goal of the schedule doesn't seem to require that a joint registry be created.

> For clarity, it might be more helpful to frame this question in terms of the specific structs in the HTTP API draft.

I'm not sure. I expect many users of Privacy Pass to just use the wire formats described in this doc.

chris-wood: > Be it PrivacyPassVOPRFP521 or PrivacyPassCoolPostQuantum or PrivacyPassBLS. Seems like something that the PrivacyPass WG should define somewhere as things that higher level things point to

@dvorak42 I guess what gives me pause is that putting everything behind a single registry implies that all values in that registry have similar properties, but that's not the case here. In contrast to, say, TLS algorithm registries, which are really just different algorithms with the same set of features. The things we're considering merging here have fundamentally different features (or at least they do in my opinion), so it's not clear to me that merging them is appropriate. (Sorry if I'm not making my thinking clear!)

bemasc: @chris-wood Maybe you could send a PR that updates draft-ietf-privacypass-http-api in the way you're thinking?

alxdavids:
Removed the IANA registry for now following conversation with @chris-wood, @dvorak42, @claucece. Also locked the OPRF ciphersuite to P384. We can figure out how to define the appropriate IANA registries in the future.

alxdavids:
@chris-wood I added the missing sections, is there anything else that needs addressing?

chris-wood:
Thanks, @alxdavids. It's good for now. We may need to update the issuance and redemption details to use the new HTTP-based authentication challenge being worked on in parallel, but the core of the issuance flow will remain the same.

alxdavids:
> Thanks, @alxdavids. It's good for now. We may need to update the issuance and redemption details to use the new HTTP-based authentication challenge being worked on in parallel, but the core of the issuance flow will remain the same.

Sounds good, shall I merge this for now?

chris-wood:
I don't think we can merge this as-is. This being a new document would need to go through the adoption process. @bemasc can say more.

chris-wood:
I did this alignment in #87. @alxdavids, please have a look!

alxdavids:
> I did this alignment in #87. @alxdavids, please have a look!

Sounds good, I will take a look at this tomorrow morning my time 👍

chris-wood:
Closing as overcome by events.

PR #92, opened 2022-02-01
Address correlation attacks

### LINKED ISSUE #90: Expand upon correlation attacks and mitigations
In the architecture document, we should have more discussion of correlation attacks and ways to mitigate. For example, interactive vs non-interactive tokens can making timing correlation attacks more or less relevant. For mitigations, the deployment model of issuer/origin (and attester) can make a difference.

chris-wood:
Drafty draft that addresses #90. I tried to generalize the problem and focus on correlation via either the _source_ for an event, like the client IP address or whatever, and the _context_ of an event, like the timestamp at which it occurs or the origin associated with the event. It is not a tremendous problem if two things are correlated via context so long as the clients are not linkable. (This is the argument with things like ODoH, where we can't link any two requests to the same client, but we can link two requests as being related.) What's most important here, I think, is source correlation, and clients can control this by working _through_ an attester or some other proxy.

What do folks think?

Starting as a draft while we work on the presentation.

Closes #90.

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -425,7 +425,53 @@ their position to try and reduce the anonymity sets that Clients belong
 to (or, user segregation). For each case, we provide mitigations that
 the Privacy Pass ecosystem must implement to prevent these actions.
 
-## Metadata Privacy Implications
+## Correlation attacks
+
+One of the requirements of the issuance protocol as described in {{issuance-protocol}}
+is that it provide unconditional input secrecy. This means that an issuance
+and its corresponding redemption events are unlinkable using the contents of
+a token. However, there are other ways in which these two events could be linked
+together, which we refer to as correlation.
+
+Correlation can problematic from a privacy perspective. To demonstrate why,
+consider the lifecycle of a cross-origin token. Clients may invoke the issuance
+protocol in one context, e.g., when visiting site foo.example, but redeem the
+token in another context, e.g., when visiting site bar.example. If an attacker
+can correlate these two events, then they know that visits to foo.example and
+bar.example are related. In contrast, per-origin token contexts are the same,
+so correlating issuance and redemption only allows an attacker to learn when
+issuance and redemption correspond to the same context.
+
+In general, each issuance and redemption event is associated with some Client
+and context. For example, an issuance event might represent Client C running
+the issuance protocol to produce a token for website foo.example at time T.
+Let cI and sI be the Client and context pair associated with an issuance event.
+Similarly, let cR and sR be the Client and context pair associated with a
+redemption event. Correlation is possible if sI = sR or cI = cR.
+
+Context correlation is possible through a number of different ways. The following
+list covers some examples:
+
+- The timestamps of issuance and redemption events for interactive tokens may be
+  sufficiently unique that an attacker can link the two by their timestamp. However,
+  even if these timestamps are not close, a small number of issuance and redemption
+  events can let an attacker link two events. For example, if there is only ever
+  a single issuance and redemption event in a day, then an attacker can link based
+  based solely on the presence of the events.
+- The origin for an issuance and redemption event can be unique for a given Client,
+  allowing an attacker to link both events.
+
+Depending on the Client identifiers used, context correlation may not constitute
+a privacy violation. In particular, because tokens are not bound to a single Client
+and therefore transferrable, an attacker cannot distinguish the scenarios where
+the same Client ran the issuance protocol and redemption protocol, or where one
+Client ran the issuance protocol and another client ran the redemption protocol.
+
+Moreover, clients can control whether or not correlation occurs by separating their
+Client identifiers presented to servers, e.g., by using an attester or other proxy
+service to relay token requests or hide their identity from the issuer (or origin).
+
+## Metadata privacy implications
 
 Any metadata bits of information can be used to further segment the
 size of the Client's anonymity set. Any Issuer that wanted to
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -429,7 +429,7 @@ the Privacy Pass ecosystem must implement to prevent these actions.
 
 One of the requirements of the issuance protocol as described in {{issuance-protocol}}
 is that it needs to provide unconditional input secrecy. This means that an issuance event
-and its corresponding redemption events are unlinkable using the contents of
+and its corresponding redemption event are unlinkable using the contents of
 a token. However, there are other ways in which these two events could be linked
 together, which we refer to as correlation.
 
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -428,7 +428,7 @@ the Privacy Pass ecosystem must implement to prevent these actions.
 ## Correlation attacks
 
 One of the requirements of the issuance protocol as described in {{issuance-protocol}}
-is that it provide unconditional input secrecy. This means that an issuance
+is that it needs to provide unconditional input secrecy. This means that an issuance event
 and its corresponding redemption events are unlinkable using the contents of
 a token. However, there are other ways in which these two events could be linked
 together, which we refer to as correlation.
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -438,7 +438,7 @@ consider the lifecycle of a cross-origin token. Clients may invoke the issuance
 protocol in one context, e.g., when visiting site foo.example, but redeem the
 token in another context, e.g., when visiting site bar.example. If an attacker
 can correlate these two events, then they know that visits to foo.example and
-bar.example are related. In contrast, per-origin token contexts are the same,
+bar.example are related. In contrast, per-origin token contexts are limited a single origin,
 so correlating issuance and redemption only allows an attacker to learn when
 issuance and redemption correspond to the same context.
 
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -433,7 +433,7 @@ and its corresponding redemption events are unlinkable using the contents of
 a token. However, there are other ways in which these two events could be linked
 together, which we refer to as correlation.
 
-Correlation can problematic from a privacy perspective. To demonstrate why,
+Correlation can be problematic from a privacy perspective. To demonstrate why,
 consider the lifecycle of a cross-origin token. Clients may invoke the issuance
 protocol in one context, e.g., when visiting site foo.example, but redeem the
 token in another context, e.g., when visiting site bar.example. If an attacker
@@ -447,7 +447,8 @@ and context. For example, an issuance event might represent Client C running
 the issuance protocol to produce a token for website foo.example at time T.
 Let cI and sI be the Client and context pair associated with an issuance event.
 Similarly, let cR and sR be the Client and context pair associated with a
-redemption event. Correlation is possible if sI = sR or cI = cR.
+redemption event. We refer to context correlation as the case where cI = cR,
+and source correlation as the case where sI = sR.
 
 Context correlation is possible through a number of different ways. The following
 list covers some examples:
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -425,7 +425,54 @@ their position to try and reduce the anonymity sets that Clients belong
...
+  a single issuance and redemption event in a day, then an attacker can link based
+  based solely on the presence of the events.
+- The origin for an issuance and redemption event can be unique for a given Client,
+  allowing an attacker to link both events.
+
+Depending on the Client identifiers used, context correlation may not constitute
+a privacy violation. In particular, because tokens are not bound to a single Client
+and therefore transferrable, an attacker cannot distinguish the scenarios where
+the same Client ran the issuance protocol and redemption protocol, or where one
+Client ran the issuance protocol and another client ran the redemption protocol.
```
bemasc, line 469: I don't think this is in scope unless we are also going to discuss architectures and protocols for (safe?) client-to-client token sharing.

draft-ietf-privacypass-architecture.md
```diff
@@ -425,7 +425,54 @@ their position to try and reduce the anonymity sets that Clients belong
...
+
+Moreover, clients can control source correlation by separating their Client identifiers
+presented to servers, e.g., by using an attester or other proxy service to relay token
+requests or hide their identity from the issuer (or origin).
```
bemasc, line 473: If the issuer doesn't know the client's identity, then it seems to me that the client doesn't need to use blind signatures at all. If this use case is in-scope, I think it would be clearer to offer a very simple Issuance Algorithm that is well-suited for it (e.g. plain ECDSA).

draft-ietf-privacypass-architecture.md
chris-wood: This is noting that clients _can_ be transferred.

draft-ietf-privacypass-architecture.md
```diff
@@ -425,7 +425,54 @@ their position to try and reduce the anonymity sets that Clients belong
...
+together, which we refer to as correlation.
+
+Correlation can be problematic from a privacy perspective. To demonstrate why,
+consider the lifecycle of a cross-origin token. Clients may invoke the issuance
+protocol in one context, e.g., when visiting site foo.example, but redeem the
+token in another context, e.g., when visiting site bar.example. If an attacker
+can correlate these two events, then they know that visits to foo.example and
+bar.example are related. In contrast, per-origin token contexts are the same,
+so correlating issuance and redemption only allows an attacker to learn when
+issuance and redemption correspond to the same context.
```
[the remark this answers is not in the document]
dvorak42: Maybe just a quick sentence about it, and then expand on it once we discuss the rate limited case where the attester has actual state that could be linked? Though maybe waiting until then is okay.

draft-ietf-privacypass-architecture.md
bemasc: I don't think we can hang our privacy safety claim on a client behavior that is never otherwise standardized, recommended, or even discussed.  Also, this kind of sharing is unsafe in general, and seems unlikely to occur in practice without substantial changes to Privacy Pass.

If you want to open this can of worms, I think you have to open it all the way.

draft-ietf-privacypass-architecture.md
bemasc: If this architecture  and protocol are intended to cover that use case, then we have an interesting charter discussion, because the charter says: "An Issuer cannot link a redeemed token to one of N previously created tokens
using the same key".

If we are covering use cases that have very small effective `N`, even `N=1`, I think that needs to be made very clear to the working group.  Defining a non-blinded token algorithm now could be a good way to focus that discussion.

draft-ietf-privacypass-architecture.md
chris-wood: I think there's a misunderstanding here. Per the charter, a goal of this protocol is to "specify an extensible protocol for creating and redeeming anonymous and transferrable tokens." The key word is _transferrable_. This paragraph is simply acknowledging that tokens _can_ in fact be transferred amongst clients -- by design! It is by no means encouraging this behavior. The can of worms is already open.

draft-ietf-privacypass-architecture.md
chris-wood: > If we are covering use cases that have very small effective N, even N=1, I think that needs to be made very clear to the working group. Defining a non-blinded token algorithm now could be a good way to focus that discussion.

I disagree. The value of N is _entirely_ deployment specific. It cannot be addressed in the protocol, without an incredibly different design.

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
chris-wood: A sentence would be good, yeah, noting that issuance protocol state can be used for linking.

draft-ietf-privacypass-architecture.md
```diff
@@ -425,7 +425,54 @@ their position to try and reduce the anonymity sets that Clients belong
 to (or, user segregation). For each case, we provide mitigations that
 the Privacy Pass ecosystem must implement to prevent these actions.
 
-## Metadata Privacy Implications
+## Correlation attacks
+
+One of the requirements of the issuance protocol as described in {{issuance-protocol}}
+is that it provide unconditional input secrecy. This means that an issuance
```
tfpauly, line 431:
```suggestion
is that it needs to provide unconditional input secrecy. This means that an issuance event
```

draft-ietf-privacypass-architecture.md
```diff
@@ -425,7 +425,54 @@ their position to try and reduce the anonymity sets that Clients belong
...
+and its corresponding redemption events are unlinkable using the contents of
```
tfpauly, line 432:
```suggestion
and its corresponding redemption event are unlinkable using the contents of
```

draft-ietf-privacypass-architecture.md
```diff
@@ -425,7 +425,54 @@ their position to try and reduce the anonymity sets that Clients belong
...
+a token. However, there are other ways in which these two events could be linked
+together, which we refer to as correlation.
+
+Correlation can be problematic from a privacy perspective. To demonstrate why,
+consider the lifecycle of a cross-origin token. Clients may invoke the issuance
+protocol in one context, e.g., when visiting site foo.example, but redeem the
+token in another context, e.g., when visiting site bar.example. If an attacker
+can correlate these two events, then they know that visits to foo.example and
+bar.example are related. In contrast, per-origin token contexts are the same,
```
tfpauly, line 441:
```suggestion
bar.example are related. In contrast, per-origin token contexts are limited a single origin,
```

draft-ietf-privacypass-architecture.md
tfpauly, line 469: How realistic is the handing-off tokens case? If it's not very common, it doesn't seem like a strong mitigation.

chris-wood: This isn't meant to suggest it as a mitigation. It's merely meant to say that an origin cannot guarantee that the same token which went through issuance is the same token which redeemed a token, because they're transferrable. (What part of the text suggests this is a mitigation?)

tfpauly: It doesn't directly say it's a mitigation, I was just inferring that.

chris-wood: Ah, okay, is there any way we could clarify? (@bemasc zeroed in on this as well, so surely the text could be improved.)

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
chris-wood: @dvorak42 added state example!

chris-wood:
@tfpauly @dvorak42 @alxdavids ready for review now. @janaiyengar, please have a look, too!

draft-ietf-privacypass-architecture.md
janaiyengar: Do we need this paragraph?

janaiyengar (commented):
One comment, else looks good!

draft-ietf-privacypass-architecture.md
janaiyengar: I had the same read as tfpauly -- that this suggests a mitigation, but you're right that it doesn't. If it's not a common use case, why use this example? And I'm then wondering if removing it entirely removes anything importnat -- it seems to me that it doesnt.

chris-wood: I think this paragraph is important because it notes that linking issuance and redemption together may not constitute an actual attack. Token transfer is a fundamental property of this protocol. This text acknowledges that property.

janaiyengar (approved):
ok, sounds good -- thanks for working on this, @chris-wood !

dvorak42 (approved):

draft-ietf-privacypass-architecture.md
bemasc: I still don't think this text is helpful.  Most digital identification systems have this transferability property.  Passwords, private keys, even hardware tokens are all transferable between people.  Transferability is trivial unless there's a reason to expect it will be unusually common.

This text seems to be arguing against the strawman position that correlation _always_ harms privacy.  In general, I don't think that there is any technical action that _always_ harms privacy.  The important question is whether there is some real use case where an otherwise-dangerous thing becomes safe.  I would prefer to see the focus there.

chris-wood: > I still don't think this text is helpful. Most digital identification systems have this transferability property. Passwords, private keys, even hardware tokens are all transferable between people. Transferability is trivial unless there's a reason to expect it will be unusually common.
> 
> This text seems to be arguing against the strawman position that correlation always harms privacy. In general, I don't think that there is any technical action that always harms privacy. The important question is whether there is some real use case where an otherwise-dangerous thing becomes safe. I would prefer to see the focus there.

I'm trying to understand you but am failing to do so. Surely there are technical actions that always harm privacy. For example, never using HTTPS is harmful, revealing your IP address and other fingerprinting information to an origin is harmful, allowing third-party cookies to track you is harmful, etc. 

Is your objection to the phrasing "privacy violation"? If not, what is it?

chris-wood: In an attempt to find a compromise, here's a proposed rephrasing of the text:

~~~
Context correlation between two events may not constitute a privacy violation if 
there is no source correlation between those two events. For example, if issuance
for a visit to foo.example was correlated to redemption of a visit to bar.example,
yet the source for these two events was uncorrelated, then one only learns that these
two events are related and might be associated with the same source (Client). In
contrast, if the source of these two events were correlated, then one learns that 
the two visits correspond to the same source (Client), thereby leaking browsing history.
~~~

@tfpauly, @janaiyengar, @bemasc, is this an improvement?

bemasc: One way to frame the problem is that this text describes a "scenario" involving multiple clients (an "acquiring client" and "redeeming client"), but this scenario is not contemplated in Section 3, which defines the scope of this design as involving only 4 communicating parties.  We shouldn't be performing security analysis based on scenarios that aren't in scope.

It seems like this section inverts the ordinary logic of security analysis.  Instead of asking "is there any strange situation where this could be dangerous?", it asks "is there any strange situation where this could be safe?".

> For example, never using HTTPS is harmful

Yes, and this text is like "Depending on the content used, insecure HTTPS may not be a privacy violation.  In particular, because the HTTP payload may already be encrypted using HPKE, ...".  This is true, but it's not helpful unless that is a relevant, reasonably common use case.

chris-wood: > One way to frame the problem is that this text describes a "scenario" involving multiple clients (an "acquiring client" and "redeeming client"), but this scenario is not contemplated in Section 3, which defines the scope of this design as involving only 4 communicating parties. We shouldn't be performing security analysis based on scenarios that aren't in scope.

I very strongly disagree with this statement! The protocol is defined in terms of 4 parties, but security and privacy need to account for multiple instances of each. This is fairly common practice.

draft-ietf-privacypass-architecture.md
bemasc: Yes, it cannot be addressed in the protocol, but it can be addressed in the architecture documentation, and already is, in section 6.  That section's analysis is invalidated by these changes, and needs to be updated.

chris-wood: Sure. Separate PR.

chris-wood:
Closing in favor of #100.

PR #99, opened 2022-03-01
Add configuration parameters

### LINKED ISSUE #96: HTTP examples
See [Mark's original issue](https://github.com/tfpauly/privacy-proxy/issues/160#issue-1125596987).

chris-wood:
Closes #96.

cc @mnot, @tfpauly

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -100,6 +100,38 @@ The following terms are used throughout this document.
 We assume that all protocol messages are encoded into raw byte format
 before being sent across the wire.
 
+# Configuration {#setup}
+
+Issuers MUST provide one parameter for configuration:
+
+1. Issuer Request URI: a token request URL for generating access tokens.
+   For example, an Issuer URL might be https://issuer.example.net/example-token-request.
+   This parameter uses resource media type "text/plain".
+
+The Issuer parameters can be obtained from an Issuer via a directory object, which is a JSON
+object whose field names and values are raw values and URLs for the parameters.
+
+| Field Name           | Value                                            |
+|:---------------------|:-------------------------------------------------|
+| issuer-request-uri   | Issuer Request URI resource URL as a JSON string |
+
+As an example, the Issuer's JSON directory could look like:
+
+~~~
+ {
+    "issuer-request-uri": "https://issuer.example.net/example-token-request"
+ }
+~~~
+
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

chris-wood:
@mnot can you help us understand what would be the preferred alternative? At the end of the day, if I'm an origin or client and need to get this information from an issuer, and all I have is the issuer name, what am I left with?

mnot:
So that's the question -- what does the person using the service have to start with? 

E.g., if I know that the service I want to work with is "cloudflare", I still have to discover something more specific, even with the current `.well-known` approach -- I need a hostname (and possibly a port). It might be `cloudflare.com`, but it equally could be `api.cloudflare.com` or `privacypass-inititate.cloudflare.com`.

So I have to find it somewhere, probably in documentation. Given that I have to go through that process as a consumer of this API, what's the advantage of configuring with just a hostname, vs. a full URL -- allowing you to skip the .well-known location?

Skipping the .well-known saves latency and also avoids some caching issues (i.e., how long can you trust the information in it for? How do you manage change?)

There are reasons that this would be the case -- e.g., if there's a lot more configuration information to convey (that hasn't made it into the spec yet), the indirection is actually desirable for operational purposes, or that the hostname _can_ be inferred / discovered automatically (e.g., with a DNS record for the registrable domain name). I'm not deep enough into this protocol to know whether any of this is the case.

chris-wood:
> E.g., if I know that the service I want to work with is "cloudflare", I still have to discover something more specific, even with the current .well-known approach -- I need a hostname (and possibly a port). It might be cloudflare.com, but it equally could be api.cloudflare.com or privacypass-inititate.cloudflare.com.

I would expect a common practice to be that folks are configured with the full authority (`issuer.example` and the implied :443 port). And [following best practices](https://www.ietf.org/id/draft-ietf-httpbis-bcp56bis-15.html#name-discovering-an-applications), they could then use to this discover additional information using a registered .well-known. 

But, to your point, if they're configured this way, it's probably reasonable to require them to simply be configured with the full URL used to obtain the configuration dictionary. That is, instead of configuring them `issuer.example`, configure them with `https://issuer.example/my-special-configuraiton-url/`. 

I'm not sure we have enough experience yet to know which of these would be preferred. It's probably worth splitting this off into a separate issue to track as things progress. @tfpauly, thoughts?

PR #100, opened 2022-03-02
Add deployment models and considerations

### LINKED ISSUE #90: Expand upon correlation attacks and mitigations

chris-wood:
Closes #90.

cc @janaiyengar, @tfpauly, @smhendrickson, @dvorak42

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -433,7 +433,8 @@ The Origin, Attester, and Issuer portrayed in {{fig-overview}} can be instantiat
 and deployed in a number of different ways. This section covers some expected
 deployment models and their corresponding security and privacy considerations.
 The discussion below assumes non-collusion between entities when operated by
-separate parties.
+separate parties. Mechanisms for enforcing non-collusion are out of scope
+for this architecture.
 
 ## Shared Origin, Attester, Issuer {#deploy-shared}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -410,15 +423,184 @@ guidelines on changes to the parametrization in {{parametrization}}.
...
+This model is useful for Origins that require Client-identifying attestation,
+e.g., through the use of application-layer account information, but do not
+otherwise want to learn information about individual Clients beyond what is
+observed during the token redemption, such as Client IP addresses.
+
+In this model, attestation and redemption contexts are separate. As a result,
+any type of attestation is suitable in this model. Moreover, any type of token
+challenge is suitable assuming Client redemption unlinkability, since no single
+party will have access to the identifying Client information and unique Origin
+information.
```
tfpauly, line 550: It may be good to point out that this model is less useful if there's only one origin for the issuer — since then the attester knows exactly what the origin is. In that case, some token types aren't advisable.

draft-ietf-privacypass-architecture.md
```diff
@@ -410,15 +423,184 @@ guidelines on changes to the parametrization in {{parametrization}}.
...
+
+Client uses Privacy Pass to separate attestation context and redemption
+context. Linking or combining these contexts can reveal sensitive information
+about the Client, including their identity or browsing history. Depending on the
+deployment model, separating these contexts can take different forms.
+The Origin, Attester, and Issuer portrayed in {{fig-overview}} can be instantiated
+and deployed in a number of different ways. This section covers some expected
+deployment models and their corresponding security and privacy considerations.
+The discussion below assumes non-collusion between entities when operated by
+separate parties.
```
tfpauly, line 436: Should we mention at all how the client knows which entities are separate? If that's out of scope, can we say so or make a reference?

tfpauly (approved):

draft-ietf-privacypass-architecture.md
chris-wood: Indeed, per-origin tokens challenged by a single origin seem useless, though I'm not sure we need to explicitly call it out. Curious to hear what others think!

draft-ietf-privacypass-architecture.md
tfpauly: Yup, that would be good to add.

draft-ietf-privacypass-architecture.md
tfpauly: I think it's important to call out, since it reduces the applicability of this deployment model. If my origin is small (only one real origin), then I should know that this probably isn't the right choice for me.

PR #107, opened 2022-03-15
Explain when to flush context-bound tokens

### LINKED ISSUE #102: Challenge and cookie cache
From: https://github.com/tfpauly/privacy-proxy/issues/105

@chris-wood said:
> Servers can offer multiple challenges in a WWW-Authenticate header to let the client fetch a pile of tokens for future spending. For tracking reasons, the lifetime of these should be bound to any cookie state associated with the origin. In particular, if the client wipes its cookies for the origin, then it should also clear any challenge state. It's possible for the server to encode client-specific information in the TokenChallenge.redemption_nonce, so we need to ensure this doesn't turn into a different type of tracking vector.

tfpauly:
Closes #102

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -197,6 +197,13 @@ WWW-Authenticate: PrivateToken challenge=abc..., token-key=123...,
 PrivateToken challenge=def..., token-key=234...
 ~~~
 
+If a Client fetches a batch of multiple tokens for future use that are bound
+to a specific redemption context (the redemption_context in the TokenChallenge
+was not empty), Clients SHOULD discard these tokens upon flushing state such as
+HTTP cookies {{?COOKIES=I-D.ietf-httpbis-rfc6265bis}}, or changing networks.
+Using these tokens in a context that otherwise would not be linkable to the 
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -197,6 +197,13 @@ WWW-Authenticate: PrivateToken challenge=abc..., token-key=123...,
 PrivateToken challenge=def..., token-key=234...
 ~~~
 
+If a Client fetches a batch of multiple tokens for future use that are bound
+to a specific redemption context (the redemption_context in the TokenChallenge
+was not empty), Clients SHOULD discard these tokens upon flushing state such as
+HTTP cookies {{?COOKIES=I-D.ietf-httpbis-rfc6265bis}}, or changing networks.
+Using these tokens in a context that otherwise would not be linkable to the 
+original context could allow the Origin to recognize a Client.
+
```
tfpauly, line 206:
```suggestion
Using these tokens in a context that otherwise would not be linkable to the
original context could allow the Origin to recognize a Client.

```

chris-wood (approved):

Viosuna (approved):

PR #111, opened 2022-03-30
Revert back to VOPRF for basic private issuance

chris-wood:
Since metadata can be bound to tokens via TokenChallenge.redemption_context, and the origin is ultimately the one who determines what metadata is supported, there's no more real need for the _issuer_ to supply this metadata. So let's go with the simpler VOPRF protocol instead!

Test vectors generated from the code here: https://github.com/cloudflare/pat-go/pull/7

cc @armfazh, @claucece, @dvorak42


armfazh (approved):
LGTM

tfpauly (approved):

PR #116, opened 2022-04-06
Add challenge and redemption test vectors

chris-wood:
We could omit the authenticator value and just focus on how Token is constructed based on TokenChallenge, but it seems reasonable to include the authenticator so folks can also test that they're computing the token key ID correctly.

Generated from [this change](https://github.com/cloudflare/pat-go/pull/8).


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -467,3 +467,203 @@ The initial contents for this registry are defined in the table below.
...
+- token_type: The type of token issuance protocol, a value from {{token-types}}.
+- issuer_name: The name of the issuer in the TokenChallenge structure, represented as a hexadecimal string.
+- redemption_context: The redemption context in the TokenChallenge structure, represented as a hexadecimal string.
+- origin_info: The origin info in the TokenChallenge structure, represented as a hexadecimal string.
+- nonce: The nonce in the Token structure, represented as a hexadecimal string.
+- token_key: The public token-key, encoded based on the corresponding token type, represented as a hexadecimal string.
+- token_authenticator_input: The values in the Token structure used to compute the Token authenticator value, represented as a hexadecimal string.
+- token_authenticator: The output Token authenticator which verifies under token_key, represented as a hexadecimal string.
+
+Test vectors are provided for each of the following TokenChallenge configurations:
```
tfpauly, line 487: Should we reference the protocol doc and specifically call out that these are for the RSA blind signature variant?

tfpauly (approved):

PR #122, opened 2022-04-22
Pin down RSA details

### LINKED ISSUE #120: Blind RSA identifier
The SPKI format is currently unspecified, and as there are multiple ways to encode an RSA public key (and hash algorithm), this makes interop somewhat difficult. Moreover, the SPKI can encode any number of RSA public key sizes, in contrast to elliptic curves wherein the "suite" typically defines the key size. In the name of simplicity, I'd propose that we have the identifier fully specify everything one needs to use the algorithm, including key size, hash algorithm, and encoding format.

chris-wood:
This is a breaking change since it updates the value of the code points. Starting as a draft until I'm able to produce matching test vectors. 

Closes #120.


tfpauly:
Hm, I was thinking we'd just fully specify the types 1 and 2 right now, rather than trying to define a bunch of different combinations up front. Is there a benefit to listing all of the combinations out? Since for interop we want convergence on types, why not just have a very standard 1 and 2?

tfpauly:
For the common web cases, where we have websites that want to work with the wide range of clients, I'd prefer to see a very clear option of what to use by default — with one for privately verifiable, and one for publicly verifiable. If people have use cases that can't use these algorithms (like we've seen for IOT use cases, etc, for other protocols), I'd rather see those defined when there's a profile that has a driving use case. For example, we could end up defining a full matrix of different SHA lengths and key sizes (they're all 4096 right now for RSA, but we could have others) and that seems unnecessary.

chris-wood:
>  If people have use cases that can't use these algorithms (like we've seen for IOT use cases, etc, for other protocols), I'd rather see those defined when there's a profile that has a driving use case.

Well, we can RECOMMEND issuance protocols for common use cases, as we've done for ECH.

draft-ietf-privacypass-protocol.md
```diff
@@ -488,10 +492,13 @@ to the Architecture document {{I-D.ietf-privacypass-architecture}}.
...
 This document updates the "Token Type" Registry with the following values.
 
-| Value  | Name                   | Publicly Verifiable | Public Metadata | Private Metadata | Nk  | Reference        |
-|:-------|:-----------------------|:--------------------|:----------------|:-----------------|:----|:-----------------|
-| 0x0001 | VOPRF(P-384, SHA-384)  | N                   | N               | N                | 48  | {{private-flow}} |
-| 0x0002 | Blind RSA, 4096        | Y                   | N               | N                | 512 | {{public-flow}}  |
+| Value  | Name                           | Publicly Verifiable | Public Metadata | Private Metadata | Nk  | Reference        |
+|:-------|:-------------------------------|:--------------------|:----------------|:-----------------|:----|:-----------------|
+| 0x0001 | VOPRF (P-384, SHA-384)         | N                   | N               | N                | 48  | {{private-flow}} |
+| 0x0001 | VOPRF (P-521, SHA-512)         | N                   | N               | N                | 64  | {{private-flow}} |
```
tfpauly, line 498: I think you mean 0x0002 here — but I'd also like to keep 2 as the default blind RSA type

chris-wood:
Reverted type2 change!

chris-wood:
@tfpauly this is good to go now -- please have a look!

tfpauly (approved):

PR #123, opened 2022-04-22
Be more specific about the key ID

### LINKED ISSUE #121: Clearly explain least significant byte in key_id
We should be explicit about what this means:

`"token_key_id" is the least significant byte of the key_id.`

This means that the byte is the last/rightmost byte of the hash, in order. Misinterpretation of this caused some confusion in interop.

chris-wood:
Closes #121.

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -380,7 +380,7 @@ The structure fields are defined as follows:
 
 - "token_type" is a 2-octet integer, which matches the type in the challenge.
 
-- "token_key_id" is the least significant byte of the `key_id` in network byte order.
+- "token_key_id" is the least significant byte of the `key_id` in network byte order (in other words, the last 8 bits of `key_id`).
 
 - "blinded_msg" is the Nk-octet request defined above.
 
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -204,7 +204,7 @@ The structure fields are defined as follows:
 
 - "token_type" is a 2-octet integer, which matches the type in the challenge.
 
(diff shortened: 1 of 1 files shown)


tfpauly:
Is this the most exact way to put this? For byte arrays like a hash, I generally don't think of it as being network- or host- byte order. Do we have other docs that refer to stuff like this that we can crib from?

chris-wood:
Nothing comes to mind right now, but I'm happy to look at alternatives.

draft-ietf-privacypass-protocol.md
```diff
@@ -204,7 +204,7 @@ The structure fields are defined as follows:
 
 - "token_type" is a 2-octet integer, which matches the type in the challenge.
 
-- "token_key_id" is the least significant byte of the `key_id`.
+- "token_key_id" is the least significant byte of the `key_id` in network byte order.
```
tfpauly, line 207:
```suggestion
- "token_key_id" is the least significant byte of the `key_id` in network byte order (in other words, the last 8 bits of `key_id`).
```

draft-ietf-privacypass-protocol.md
```diff
@@ -380,7 +380,7 @@ The structure fields are defined as follows:
...
```
tfpauly, line 383:
```suggestion
- "token_key_id" is the least significant byte of the `key_id` in network byte order (in other words, the last 8 bits of `key_id`).
```

tfpauly (commented):
Here's an attempt at saying this in other words to be extra clear?

draft-ietf-privacypass-protocol.md
chris-wood: 👍 Great!

chris-wood:
Works for me =)

PR #127, opened 2022-06-14
Advertise token-key values in the issuer configuration dictionary

chris-wood:
This is both a convenience for origins, as well as a way to support key rotation.

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -125,7 +125,9 @@ Each "token-keys" JSON object contains the following fields and corresponding ra
 | token-key    | The base64url encoding of the public key for use with the issuance protocol, including padding, as a JSON string |
 
 Issuers MAY advertise multiple token-keys for the same token-type to
-support key rotation.
+support key rotation. In this case, Issuers indicate preference for which
+token key to use based on the order of keys in the list, with preference
+given to keys earlier in the list.
 
 Altogether, the Issuer's JSON directory could look like:
 
@@ -136,12 +138,10 @@ Altogether, the Issuer's JSON directory could look like:
       {
         "token-type": 2,
         "token-key": "MI...AB",
-        "version": 1,
       },
       {
         "token-type": 2,
         "token-key": "MI...AQ",
-        "version": 2,
       }
     ]
  }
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -102,24 +102,48 @@ before being sent across the wire.
...
 
 ~~~
  {
-    "issuer-request-uri": "https://issuer.example.net/example-token-request"
+    "issuer-request-uri": "https://issuer.example.net/example-token-request",
+    "token-keys": [
+      {
+        "token-type": 2,
+        "token-key": "MI...AB",
+        "version": 1,
```
tfpauly, line 139: What's the version field?

chris-wood: Moved to order-based preference.

tfpauly (approved):

chris-wood:
> IIUC the key rotation interval for non-rate-limited token issuance may be ambiguous. Since there is not necessarily a policy window defined in that case. Is that right?

Correct. The rotation window is currently not specified. Omitting that seems reasonable for this case. Origins can simply check in with issuers at some regular cadence as they deem fit.

draft-ietf-privacypass-protocol.md
```diff
@@ -102,24 +102,48 @@ before being sent across the wire.
...
+    "token-keys": [
+      {
+        "token-type": 2,
+        "token-key": "MI...AB",
+      },
+      {
+        "token-type": 2,
+        "token-key": "MI...AQ",
+      }
+    ]
```
armfazh, line 146:
```suggestion
    "issuance-keys": [
      {
        "token-type": 2,
        "token-keys": [ "MI...AB", "MI...AQ" ],
      },
      {
        "token-type": 3,
        "token-keys": ["MI...AW"],
      }
    ]
```

May I suggest to group keys by token-type? The advantages are at parsing this information.

armfazh (approved):

draft-ietf-privacypass-protocol.md
chris-wood: Hmm, how does this help with parsing? With the current scheme, one can implement a function like `CreateTokenKeyFromJSON(...)` using each token-key object. With the proposal, that would be something like `CreateTokenKeyListFromJSON(...)`, returning a list of items. The former seems like it'd yield simpler implementations. 

@jfoote, thoughts?

jfoote: I agree with @chris-wood.

PR #131, opened 2022-06-27
Public means public after all

### LINKED ISSUE #118: Explain consequences of using publicly verifiable token
The publicly verifiable basic token issuance protocol doesn't include an origin name that the issuer can see, so it does allow origins to be able to use that issuer (as long as the know the key) without explicit coordination or permission from the issuer. This isn't necessarily a problem, but we should likely mention it.

chris-wood:
Closes #118.

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -360,8 +360,8 @@ in two important ways:
 1. The issuance protocol does not admit public or private metadata to bind
    additional context to tokens.
 
-The former property means that any Origin can select the Issuer to produce tokens provided
-the Origin has the Issuer public key, without explicit coordination or permission
+The first property means that any Origin can select a given Issuer to produce tokens, as long
+as the Origin has the Issuer public key, without explicit coordination or permission
 from the Issuer. This is because the Issuer does not learn the Origin that requested
 the token during the issuance protocol.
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -356,13 +356,19 @@ for producing publicly verifiable tokens. It differs from the previous variant
...
 1. The issuance protocol does not admit public or private metadata to bind
    additional context to tokens.
 
-Otherwise, this variant is nearly identical. In particular, Issuers provide a
-Private and Public Key, denoted skI and pkI, respectively, used to produce tokens
-as input to the protocol. See {{public-issuer-configuration}} for how this key
+The former property means that any Origin can select the Issuer to produce tokens provided
+the Origin has the Issuer public key, without explicit coordination or permission
+from the Issuer. This is because the Issuer does not learn the Origin that requested
+the token during the issuance protocol.
```
tfpauly, line 366:
```suggestion
The first property means that any Origin can select a given Issuer to produce tokens, as long
as the Origin has the Issuer public key, without explicit coordination or permission
from the Issuer. This is because the Issuer does not learn the Origin that requested
the token during the issuance protocol.
```

tfpauly (approved):

PR #132, opened 2022-06-27
Remove parameterization section as it's redundant with other consideration sections

### LINKED ISSUE #65: Update Privacy Calculations in Architecture Document
The privacy calculations should be updated to take into account metadata more explicitly, and take into account the cost of the number of redemptions, rather than the global issuer count.

If used in a multiple disjoint architecture ecosystem, then the math should also be taken in terms of a specific 'context' rather than treating the whole PrivacyPass ecosystem as a single global system (allowing for more nuanced requirements/limitations on issuer counts).

chris-wood:
Closes #65.

After quickly skimming the parameterization section, it became clear that we already discuss these things in various consideration sections throughout the document, so this was just redundant. Moreover, it's very use-case and deployment specific that once had value in the joint deployment model, so it doesn't seem like particularly helpful now that we've generalized things a bit. Removing and avoiding confusion seems like the best path forward.

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -265,8 +265,8 @@ with the Issuer directly.
 with arbitrarily many Clients.
 
 Each Issuance protocol MUST come with a detailed analysis of the privacy impacts
-of the protocol, why these impacts are justified, and guidelines on changes to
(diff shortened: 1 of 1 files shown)


chris-wood:
I'll take this one to the list before merging.

PR #144, opened 2022-07-22
Add IANA registry review details

### LINKED ISSUE #143: IANA registry details for token types
From IANA:

> First, where should this new registry be located? Should it be added to an existing registry page? If it needs a new page, does it also need a new category at http://www.iana.org/protocols (and if so, should the page and the category have the same name)?
> 
> Second, what is the registration procedure for this new registry? If you need more information about common procedures, see Section 4 of RFC 8126:
> 
> https://datatracker.ietf.org/doc/html/rfc8126#section-4

tfpauly:
Closes #143

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -461,7 +461,13 @@ Template:
 
 New entries in this registry are subject to the Specification Required registration policy
 ({{!RFC8126, Section 4.6}}). Designated experts need to ensure that the token type
-is sufficiently clearly defined to be used for both token issuance and redemption.
+is sufficiently clearly defined to be used for both token issuance and redemption,
+and meets the common security requirements for issuance protocols defined
+in {{Section 3.2 of ARCHITECTURE}}.
+
+This registry also will also allow provisional registrations to allow for experimentation
+with protocols being developed. Designated experts review and approve provisional
+registrations.
 
 This document defines several Reserved values, which can be used by clients and servers
 to send "greased" values in token challenges and responses to ensure that implementations
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -462,11 +462,11 @@ Template:
 New entries in this registry are subject to the Specification Required registration policy
 ({{!RFC8126, Section 4.6}}). Designated experts need to ensure that the token type
 is sufficiently clearly defined to be used for both token issuance and redemption,
-and meets the common security requirements for issuance protocols defined
+and meets the common security and privacy requirements for issuance protocols defined
 in {{Section 3.2 of ARCHITECTURE}}.
 
 This registry also will also allow provisional registrations to allow for experimentation
-with protocols being developed. Designated experts review and approve provisional
+with protocols being developed. Designated experts review, approve, and revoke provisional
 registrations.
 
 This document defines several Reserved values, which can be used by clients and servers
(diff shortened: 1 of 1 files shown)


chris-wood (approved):

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -458,6 +459,10 @@ Template:
 * Nid: The length of the token key identifier
 * Reference: Where this algorithm is defined
 
+New entries in this registry are subject to the Specification Required registration policy
+({{!RFC8126, Section 4.6}}). Designated experts need to ensure that the token type
+is sufficiently clearly defined to be used for both token issuance and redemption.
```
tfpauly, line 464:
```suggestion
is sufficiently clearly defined to be used for both token issuance and redemption,
and meets the common security requirements for issuance protocols defined
in {{Section 3.2 of ARCHITECTURE}}.

This registry also will also allow provisional registrations to allow for experimentation
with protocols being developed. Designated experts review and approve provisional
registrations.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -458,6 +459,16 @@ Template:
...
+New entries in this registry are subject to the Specification Required registration policy
+({{!RFC8126, Section 4.6}}). Designated experts need to ensure that the token type
+is sufficiently clearly defined to be used for both token issuance and redemption,
+and meets the common security requirements for issuance protocols defined
+in {{Section 3.2 of ARCHITECTURE}}.
+
+This registry also will also allow provisional registrations to allow for experimentation
+with protocols being developed. Designated experts review and approve provisional
+registrations.
+
```
tfpauly, line 471: @chris-wood please check this new text I added based on the IETF meeting discussion

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -458,6 +459,16 @@ Template:
 * Nid: The length of the token key identifier
 * Reference: Where this algorithm is defined
 
+New entries in this registry are subject to the Specification Required registration policy
+({{!RFC8126, Section 4.6}}). Designated experts need to ensure that the token type
+is sufficiently clearly defined to be used for both token issuance and redemption,
+and meets the common security requirements for issuance protocols defined
```
chris-wood, line 465:
```suggestion
and meets the common security and privacy requirements for issuance protocols defined
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -458,6 +459,16 @@ Template:
...
+in {{Section 3.2 of ARCHITECTURE}}.
+
+This registry also will also allow provisional registrations to allow for experimentation
+with protocols being developed. Designated experts review and approve provisional
+registrations.
```
chris-wood, line 470:
```suggestion
with protocols being developed. Designated experts review, approve, and revoke provisional
registrations.
```

draft-ietf-privacypass-auth-scheme.md
chris-wood: So the idea is to leave the experimental range (and duration) to the experts? If so, that works for me. I suggested a small edit that allows them to revoke provisional status so that we don't exhaust the space with experimental values.

chris-wood (approved):
LGTM pending suggestions.

draft-ietf-privacypass-auth-scheme.md
tfpauly: I don't think it's necessarily a strict experimental range, but rather allows for early allocation that then can expire and get replaced eventually.

draft-ietf-privacypass-auth-scheme.md
tfpauly: Yeah, that works!

PR #147, opened 2022-08-01
Add text around client and origin ossification

### LINKED ISSUE #146: Preventing reliance on privacy pass
From IETF 114 discussion:

For the general captcha use cases, we don't want websites to rely on privacy pass always working, to preventing locking out other clients.

We should likely have discussion in the architecture document to suggest approaches to make sure servers can't rely on this for generic clients (like browsers). Randomly not responding with tokens, other client-side heuristics.

chris-wood:
Closes #146.

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -383,7 +383,7 @@ Origins SHOULD account for such operational or interoperability failures by offe
 clients an alternative type of challenge such as CAPTCHA for accessing a resource.
 
 To mitigate the risk of deployments becoming dependent on tokens, clients and servers SHOULD grease
-their behavior unless otherwise configured to do so. In particular, clients SHOULD
+their behavior unless explicitly configured not to. In particular, clients SHOULD
 ignore token challenges with some non-zero probability. Likewise, origins SHOULD
 randomly choose to not challenge clients for tokens with some non-zero probability.
 Moreover, origins SHOULD include random token types in token challenges with some
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -378,7 +378,7 @@ or a different token type, to allow clients to choose a preferred issuer or type
 
 An origin MUST NOT assume that token challenges will always yield a valid token.
 Clients might experience issues running the issuance protocol, e.g., because the attester
-or issuer is unavailable, or clients may simply not support the requested token type.
+or issuer is unavailable, or clients might simply not support the requested token type.
 Origins SHOULD account for such operational or interoperability failures by offering
 clients an alternative type of challenge such as CAPTCHA for accessing a resource.
 
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -382,7 +382,7 @@ or issuer is unavailable, or clients might simply not support the requested toke
 Origins SHOULD account for such operational or interoperability failures by offering
 clients an alternative type of challenge such as CAPTCHA for accessing a resource.
 
-To mitigate ossification around token dependencies, clients and servers SHOULD grease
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -376,6 +376,19 @@ the same request and SHOULD alert the user if possible.
 Origins MAY include multiple challenges, where each challenge refers to a different issuer
 or a different token type, to allow clients to choose a preferred issuer or type.
 
+An origin MUST NOT assume that token challenges will always yield a valid token.
+Clients may experience issues running the issuance protocol, e.g., because the attester
```
tfpauly, line 380:
```suggestion
Clients might experience issues running the issuance protocol, e.g., because the attester
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -376,6 +376,19 @@ the same request and SHOULD alert the user if possible.
...
+or issuer is unavailable, or clients may simply not support the requested token type.
```
tfpauly, line 381:
```suggestion
or issuer is unavailable, or clients might simply not support the requested token type.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -376,6 +376,19 @@ the same request and SHOULD alert the user if possible.
...
+Origins SHOULD account for such operational or interoperability failures by offering
+clients an alternative type of challenge such as CAPTCHA for accessing a resource.
+
+To mitigate ossification around token dependencies, clients and servers SHOULD grease
```
tfpauly, line 385:
```suggestion
To mitigate the risk of deployments becoming dependent on tokens, clients and servers SHOULD grease
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -376,6 +376,19 @@ the same request and SHOULD alert the user if possible.
...
+their behavior unless otherwise configured to do so. In particular, clients SHOULD
```
tfpauly, line 386:
```suggestion
their behavior unless explicitly configured not to. In particular, clients SHOULD
```

tfpauly (approved):
Editorial suggestions

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -376,6 +376,19 @@ the same request and SHOULD alert the user if possible.
 Origins MAY include multiple challenges, where each challenge refers to a different issuer
 or a different token type, to allow clients to choose a preferred issuer or type.
 
+An origin MUST NOT assume that token challenges will always yield a valid token.
+Clients might experience issues running the issuance protocol, e.g., because the attester
+or issuer is unavailable, or clients might simply not support the requested token type.
+Origins SHOULD account for such operational or interoperability failures by offering
+clients an alternative type of challenge such as CAPTCHA for accessing a resource.
+
+To mitigate the risk of deployments becoming dependent on tokens, clients and servers SHOULD grease
```
dvorak42, line 385: Should there be a recommendation for attesters to abort the issuance protocol with some non-zero probability?

chris-wood: This is an interesting recommendation. For consistency reasons I could see it working, but it also might make tracking down issues that are isolated to either attester or issuer difficult in production. @tfpauly, thoughts?

tfpauly: I could see it working as an option. To some degree, it's out of scope for this auth scheme document, since we only know about origins and clients here. I'd probably leave it off for now?

chris-wood: Yeah, I forgot this was in the auth scheme doc, which knows nothing about attesters or issuers.

draft-ietf-privacypass-auth-scheme.md
tahaaden: hi

PR #148, opened 2022-08-01
Add text on attestation compromise

### LINKED ISSUE #141: Feedback on compromised attestations
Over a long enough timeline, attestations will be broken - through willful post-boot compromise, malware, or other means. Services that depend on attestations (consumers of access tokens) will observe clients that appear to pass attestation, but engage in behavior that indicates that the device is compromised (e.g. trigger thousands of interactions per second).

As it stands, the service consuming access tokens is not able to provide feedback that these attestations appear to be false negatives, and should be re-examined by the attester. Even if this feedback were to be provided out of band, it will be challenging for the attester to identify which devices are compromised and diagnose the nature of the compromise.

I do not anticipate that this will become a pressing issue until the initial attestations are compromised, but it may be helpful to plan ahead for this scenario.

chris-wood:
Closes #141.

cc @janaiyengar

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -327,6 +327,16 @@ device is lesser than the number of Clients that have solved a CAPTCHA in the
 past day. Attesters should not admit attestation types that result in small
 anonymity sets.
 
+The trustworthiness of attesters depends on their ability to correctly and
+reliably perform attestation during the issuance protocol. However, in practice,
+certain types of attestation can vary in value over time, e.g., if the
+attestation process is compromised or maliciously automated. These are
+considered exceptional events and require configuration changes to address
+the underlying cause. For example, if attestation is compromised because
+of a zero-day exploit on compliant devices, then the corresponding attestation
+format should be untrusted until the exploit is patched. Addressing changes in
+attestation quality is therefore a deployment specific task.
+
 ### Issuer Role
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -327,6 +327,17 @@ device is lesser than the number of Clients that have solved a CAPTCHA in the
...
+reliably perform attestation during the issuance protocol. However, in practice,
+certain types of attestation can vary in value over time, e.g., if the
+attestation process is compromised or maliciously automated. These are
+considered exceptional events and require configuration changes to address
+the underlying cause. For example, if attestation is compromised because
+of a zero-day exploit on compliant devices, then the corresponding attestation
+format should be untrusted until the exploit is patched. Addressing changes in
+attestation quality is therefore a deployment-specific task. In Split Attester and Issuer deployments, Issuers can choose to remove compromised Attesters from their trusted
+set until the compromise is patched, without needing to modify Origin allow-lists. 
+
```
tfpauly, line 340:
```suggestion
attestation quality is therefore a deployment-specific task. In Split Attester and
Issuer deployments, Issuers can choose to remove compromised Attesters from their trusted
set until the compromise is patched, without needing to modify Origin allow-lists.

```

dvorak42 (approved):

PR #164, opened 2022-09-06
Editorial nits found during WGLC review

ekinnear:
Entirely editorial nits found during my WGLC review that were too minor for an issue. 

Feel free to take/reject any of these, they're purely editorial and I'm not attached to any of them. 

I tried to not touch any line-length outside of paragraphs that I was editing, but I did re-wrap anything that I caused to be different enough to be an issue.


chris-wood (approved):

chris-wood:
Thanks!

PR #178, opened 2022-09-14
Address WGLC review issues

### LINKED ISSUE #153: Architecture: One protocol or two?
The intro says this:

> Privacy Pass is a protocol for authorization based on anonymous-credential authentication mechanisms. 

It also says this:

> At a high level, Privacy Pass is composed of two protocols: issuance and redemption.

It may be clearer to not call Privacy Pass both "a protocol" and "two protocols". Either we call issuance and redemption "sub-protocols", or we call the main thing "a mechanism", "an architecture", or "a set of protocols". I kinda lean towards the last option.

### LINKED ISSUE #154: Architecture: Origin definition
Origin is defined currently as:

> Origin: An entity that challenges Clients for tokens.

I wonder if it may be better to say that the Origin is fundamentally the entity that redeems tokens presented by Clients (and can challenge them too).

### LINKED ISSUE #155: Cross-origin MUST NOT
> If tokens protected with resources are unique to a single Origin, then said Origin MUST NOT admit cross-origin tokens for authorization.

I'm not sure I follow what this is saying. Does it intend to say "resources protected with tokens"? It also seems like a hard thing to enforce normatively without being more clear.

### LINKED ISSUE #156: Architecture: Forward reference for split model, etc.
Section 3.2.1 has this:

> In Split Attester and Issuer deployments, Issuers can choose to remove compromised Attesters from their trusted set until the compromise is patched, without needing to modify Origin allow-lists.

It may be good to have a reference to Section 4 about what the split deployment model is. Potentially useful in more parts of 3.2 as well.

### LINKED ISSUE #158: Architecture: Who does what and definition of "Client" in the introduction
In the introduction, client is capitalized for the first time here: 
> The issuance protocol runs between a Client and two network functions in the Privacy Pass architecture: Attestation and Issuance. 

despite having been used in the proceeding paragraphs to help illustrate the overall goals of Privacy Pass. I'm assuming that's intentional as it transitions from colloquial use of "client" to the formally defined "Client", perhaps that should be made clear, something like: 
> The issuance protocol runs between an endpoint acting in the role of a Client and two network functions in the Privacy Pass architecture: Attestation and Issuance. 

or
> The issuance protocol runs between an endpoint referred to as a Client and two network functions in the Privacy Pass architecture: Attestation and Issuance. 

Perhaps more importantly, how is it that `issuance` runs between an endpoint (Client) and a "network function" described as one of Attestation and Issuance? I kind of expected issuance to run between a `Client` and `Some other entity` (Issuer?) for the _purposes_ of issuance, not attestation.

Later on, this proves true via Issuer/Attester, so perhaps worth updating the introduction to match.

This may just be fallout of having a structure like so: 
```
Issuance -> Attestation, Issuance
Redemption -> Redemption (only?)
```
Likely solvable via some light editorial tweaks in the introduction, but also wanted to note that it might be worth just presenting it as a single flat layer: `Issuance`, `Attestation`, and `Redemption`, where the Issuer and Attester might be the same entity and the Origin is involved for Redemption. Could simplify things, and that seems to match how the rest of the document handles it.

### LINKED ISSUE #159: Architecture: Consistency in order and diagrams for redemption protocol vs. issuance protocol
A minor nit, but it was a little odd leaving the abstract/introduction which was very focused on Issue/Attest _then_ Redemption to have 3.1 be the Redemption protocol in reverse order. Is it possible to just move text around in the Introduction so that it's all consistent? 

This could also potentially be a nice moment to reconcile the diagram in the introduction: 
```
    Origin          Client        Attester          Issuer
  /--------------------------------------------------------------------
  |                 /-----------------------------------------\
  |   Challenge ----> Attest --->                             |
  |                 | TokenRequest --------------->           |
  |   Redemption    |                              (validate) | Issuance
  |      Flow       |                              (evaluate) |   Flow
  |                 |     <-------------------  TokenResponse |
  |   <--- Response |                                         |
  |                 \-----------------------------------------/
  \--------------------------------------------------------------------
```
with the one in 3.2:
```
  Origin          Client        Attester          Issuer

                  +--------------------------------------\
    Challenge ----> TokenRequest --->                    |
                  |             (attest)                 |
                  |                TokenRequest --->     |
                  |                            (evaluate)|
                  |                   <--- TokenResponse |
      Token  <----+ TokenResponse <---                   |
                  |--------------------------------------/
```

Notably, `Attest` is under Client in one but under Attester in the other and the token response -> token flow is subtly different, although it would appear that's not intentionally conveying a difference in meaning.

### LINKED ISSUE #160: Architecture nit: RFC 2119 keywords in section 3
Nit: Section 3 contains quite a few places where 2119 keywords are used uncapitalized, which can be fine, but there are enough of them that it might be worth doing a quick pass to make sure that's all intentional. 

For example: 

> The fundamental requirement for key management and discovery is that Issuers must be unable to target specific clients with unique keys without detection.

which was proceeded by: 

> To facilitate issuance, the Issuer MUST hold an Issuance key pair at any given time. The Issuer public key MUST be made available to all Clients in such a way that key rotations and other updates are publicly visible. 

All three of these requirements seem rather equally important, and while normative text doesn't need to use capitalized 2119 keywords, I'm assuming there wasn't intentionally any significance to the difference in their use between those paragraphs.

### LINKED ISSUE #161: Architecture: Extension/issuer protocol requirements for privacy impact analysis
Section 3.2 states: 
> Each Issuance protocol MUST include a detailed analysis of the privacy impacts
of the protocol, why these impacts are justified, and guidelines on how to
deploy the protocol to minimize any privacy impacts.

However, Section 3.2.4 states: 
> Each issuance protocol SHOULD include a detailed analysis of the
privacy impacts of the extension, why these impacts are justified, and
guidelines on how to deploy the protocol to minimize any privacy impacts.

_(After editorial nits are resolved in each section)_

This likely should be consistent everywhere, or even just left to the explicit extensibility section (3.2.4) and referenced from 3.2.

### LINKED ISSUE #162: Architecture: Key rotation recommendations don't depend on number and frequency of client interactions?
Section 5.2 states:
> With a large number of Clients, a minimum of one week gives a large enough
window for Clients to participate in the issuance protocol and thus enjoy the
anonymity guarantees of being part of a larger group. A maximum of
12 weeks limits the damage caused by a key compromise. If an Issuer
realizes that a key compromise has occurred then the Issuer should
generate a new key and make it available to Clients. If possible, it should
invoke any revocation procedures that may apply for the old key.

Is this 1-12 week recommendation really universally applicable, or should an implementor/deployer take into account the number and frequency of client interactions? It seems like the real goal is the _size_ of the anonymity set, not the number of days (modulo risk of key compromise). For example, if there's a system that only has 10 clients a week, would it make sense to use a longer value?

### LINKED ISSUE #165: Arch: s.3.1: Meaning of "cross-origin"
> For example, Origins that accept cross-origin across shared redemption contexts tokens SHOULD track which tokens have been redeemed already in those redemption contexts, since these tokens can be issued and then spent multiple times in response to any such challenge.

I'm not sure what "cross-origin" means in this context. (I am not an HTTP expert,) "Cross-origin token" is used in the next paragraph. Is that what's meant here?

EDIT: After reading more of the draft, it sounds like "cross-origin token" is indeed what's meant. However I'm not sure if this is ever clearly defined. Is a cross-origin token one that can be redeemed by multiple origins?

### LINKED ISSUE #166: Arch: s.3.2: Concurrent security
> 3. Concurrent security. The issuance protocol MUST be safe to run concurrently with arbitrarily many Clients.

If I understand correctly, there may be a many-to-one relationship between Attesters and Issuers. Is that right? If so, it might be worth refining. (I would also say there might be many Issuers, as there might be a many-to-one relationship between Clients and Issuers as well.)

> 3. Concurrent security. The issuance protocol MUST be safe to run concurrently with arbitrarily many Clients, Attesters and Issuers.

### LINKED ISSUE #167: Arch: s.3.2.1: SHOULD versus should
The keyword SHOULD is used in this document to suggest considerations for PP deployments. These kinds of considerations are also mentioned in this section, but w/o SHOULD. For example:

> Attesters should not admit attestation types that result in small anonymity sets.

This should could be SHOULD, I think. (Sorry for the nightmare of a sentence that was.) Similarly here:

> In general, Attesters should accept a limited form of attestation formats.

If this is important (perhaps not?) it might be worth going through the rest of the document to check for consistency.

### LINKED ISSUE #168: Arch: Nit: Inconsistent capitalization
Sometimes "Issuance protocol" is written, other times it reads "issuance protocol". (Similarly for "cross-Origin".) Which is preferred? Is consistency a concern?

### LINKED ISSUE #169: Arch: s.5.3: Meaning of "distribution of Issuer trust"
I'm not sure how to interpret the following:

> If the distribution of Issuer trust is anything close to a uniform distribution, then this is likely to uniquely identify any Client amongst all other Internet users. Assuming a uniform distribution is clearly the worst-case scenario, and unlikely to be accurate, but it provides a stark warning against allowing too many Issuers at any one time.

What is meant the "distribution of Issuer trust" here?

Relatedly, can you explain the math here:

> We propose that allowing no more than 4 Issuers at any one time is highly preferable (leading to a maximum of 64 possible user segregations).

My understanding is that the set of users can be partitioned into 2^4 = 16 different sets (not 64). More than likely I'm missing something obvious. Either way, this seems like a reasonably conservative recommendation. (It's hard for me to imagine more than a handful of Issuers being active in a deployment.)

### LINKED ISSUE #170: Arch: s.5.3.1: Who is the verifier
> As long as each redemption partition maintains a strong privacy boundary with each other, the verifier will only be able to learn a number of bits of information up to the limits within that "redemption partitions".

It's not clear from context (to me, at least) who the verifier is here. I suppose the Origin? If so, I'd suggest replacing "verifier" with "Origin".

Similar question for s.6.1.

### LINKED ISSUE #171: Arch: s.6.1: Threat model for Token exhaustion
The start of s.6 states:

> We present [...] security considerations that prevent malicious Clients from abusing the protocol.

Then the Token exhaustion refers to a "malicious verifier" who can force a client into spending tokens:

> When a Client holds tokens for an Issuer, it is possible for any verifier to invoke that client to redeem tokens for that Issuer. This can lead to an attack where a malicious verifier can force a Client to spend all of their tokens from a given Issuer. 

I'm not sure who the "verifier" is. (See #170. As far as I can tell, s.5.3 and s.6.1 are the only places that the term "verifier" is used.) In any case, I would suggest reconciling the "malicious Client" with the "malicious verifier" who forces the Client into spending tokens. As is, the threat model is a bit unclear, at least to me.

### LINKED ISSUE #173: What is an 'architectural framework'?
I think this document could be titled

> Privacy Pass Architecture

or 

> Privacy Pass Framework

... but I'm not sure that "Privacy Pass Architectural Framework" is called for...

### LINKED ISSUE #174: Arch: Relationship to HTTP
I _think_ this protocol is not specific to HTTP. If that's the case, the use of 'cross-origin' before the defintion of 'Origin' (as well as the difference in case) might confuse some readers, especially those familiar with the Web.

### LINKED ISSUE #175: Arch: Abstract is very... abstract.
The abstract doesn't really reveal what the document is doing. I found this statement near the end of the introduction more useful:

> This document describes requirements for both issuance and redemption
   protocols.  This document also describes ecosystem considerations
   that impact the stated privacy and security guarantees of the
   protocol.

chris-wood:
Closes #175.
Closes #174.
Closes #173.
Closes #171.
Closes #170.
Closes #169.
Closes #168.
Closes #167.
Closes #166.
Closes #165.
Closes #162.
Closes #161.
Closes #160.
Closes #159.
Closes #158.
Closes #156.
Closes #155.
Closes #154.
Closes #153.

cc @mnot, @tfpauly, @ekinnear, @cjpatton

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -259,7 +259,7 @@ from the issuance flow.
 Clients or Attesters (acting as Clients) to forge tokens offline or otherwise
 without interacting with the Issuer directly.
 1. Concurrent security. The issuance protocol MUST be safe to run concurrently
-with arbitrarily many Clients.
+with arbitrarily many Clients, Attesters and Issuers.
 
 Each issuance protocol MUST include a detailed analysis of the privacy impacts
 of the protocol, why these impacts are justified, and guidelines on how to
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -262,9 +262,8 @@ without interacting with the Issuer directly.
 1. Concurrent security. The issuance protocol MUST be safe to run concurrently
 with arbitrarily many Clients, Attesters and Issuers.
 
-Each issuance protocol MUST include a detailed analysis of the privacy impacts
-of the protocol, why these impacts are justified, and guidelines on how to
-deploy the protocol to minimize any privacy impacts.
+See {{extensions}} for requirements on new issuance protocol variants and related
+extensions.
 
 Clients obtain the Issuer public key directly from the Origin using the process
 described in {{HTTP-Authentication}}. Clients MAY apply some form of key
@@ -425,7 +424,7 @@ cryptographic protocol.
 
 The Privacy Pass architecture and ecosystem are both intended to be receptive to
 extensions that expand the current set of functionalities through new issuance
-protocols. Each issuance protocol SHOULD include a detailed analysis of the
+protocols. Each issuance protocol MUST include a detailed analysis of the
 privacy impacts of the extension, why these impacts are justified, and
 guidelines on how to deploy the protocol to minimize any privacy impacts.
 Any extension to the Privacy Pass protocol MUST adhere to the guidelines
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -99,12 +99,14 @@ These proofs, or tokens, are anonymous in the sense that a given token cannot
 be linked to the protocol instance in which that token was initially issued.
 
 At a high level, the Privacy Pass architecture consists of two protocols:
-issuance and redemption. The issuance protocol runs between a Client and two network functions in the
-Privacy Pass architecture: Attestation and Issuance. These two network
-functions can be implemented by the same protocol participant, but can also be
-implemented separately. The Issuer is responsible for issuing tokens in
-response to requests from Clients. The Attester is responsible for attesting to
-properties about the Client for which tokens are issued. The Issuer needs to be
+issuance and redemption. The issuance protocol runs between an endpoint referred
+to as a Client and two functions in the Privacy Pass architecture:
+Attestation and Issuance. These two network functions can be implemented by the
+same protocol participant, but can also be implemented separately. The entity that
+implements Issuance, referred to as the Issuer, is responsible for issuing tokens in
+response to requests from Clients. The entity that implements Attestation, referred to
+as the Attester, is responsible for attesting to properties about
+the Client for which tokens are issued. The Issuer needs to be
 trusted by the server that later redeems the token. Attestation can be
 performed by the Issuer or by an Attester that is trusted by the Issuer.
 Clients might prefer to select different Attesters, separate from the Issuer,
@@ -118,10 +120,9 @@ CAPTCHA and the resulting token produced by the Issuer was proof of this fact.
 
 The redemption protocol runs between Client and Origin (server). It allows
 Origins to challenge Clients to present one or more tokens for authorization.
-Depending on the type of token, e.g., whether or not it is cross-origin
-or per-origin, and whether or not it can be cached, the Client either presents
-a previously obtained token or invokes the issuance protocol to acquire one
-for authorization.
+Depending on the type of token, e.g., whether or not it can be cached, the
+Client either presents a previously obtained token or invokes the issuance
+protocol to acquire one for authorization.
 
 The issuance and redemption protocols operate in concert as shown in
 the figure below.
@@ -197,20 +198,21 @@ there are a number of ways in which the token may vary, including:
   non-interactively, whereas a fresh and random redemption context means
   that the redeemed token must be issued only after the client receives the challenge.
   See Section 2.1.1 of {{HTTP-Authentication}} for more details.
-- Per-origin or cross-origin. Tokens can be constrained to the Origin for
-  which the challenge originated, or can be used across Origins.
+- Per-Origin or cross-Origin. Tokens can be constrained to the Origin for
+  which the challenge originated (referred to as per-Origin tokens), or
+  can be used across multiple Origins (referred to as cross-Origin tokens).
 
 Depending on the use case, Origins may need to maintain state to track
-redeemed tokens. For example, Origins that accept cross-origin
+redeemed tokens. For example, Origins that accept cross-Origin
 across shared redemption contexts tokens SHOULD track which tokens
 have been redeemed already in those redemption contexts, since these
 tokens can be issued and then spent multiple times in response to any
 such challenge. See Section 2.1.1 of {{HTTP-Authentication}} for discussion.
 
-Origins that admit cross-origin tokens bear some risk of allowing tokens
+Origins that admit cross-Origin tokens bear some risk of allowing tokens
 issued for one Origin to be spent in an interaction with another Origin.
 If tokens protected with resources are unique to a single Origin, then
-said Origin MUST NOT admit cross-origin tokens for authorization.
+said Origin MUST NOT admit cross-Origin tokens for authorization.
 
 ## Issuance Protocol
 
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -204,6 +204,8 @@ there are a number of ways in which the token may vary, including:
 - Per-Origin or cross-Origin. Tokens can be constrained to the Origin for
   which the challenge originated (referred to as per-Origin tokens), or
   can be used across multiple Origins (referred to as cross-Origin tokens).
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

chris-wood:
I should have stacked PRs, but these issues were small enough to address that I just threw them into separate commits here. Sorry, @tfpauly!

draft-ietf-privacypass-architecture.md
```diff
@@ -203,20 +201,20 @@ there are a number of ways in which the token may vary, including:
...
-redeemed tokens. For example, Origins that accept cross-origin
-across shared redemption contexts tokens SHOULD track which tokens
-have been redeemed already in those redemption contexts, since these
-tokens can be issued and then spent multiple times in response to any
-such challenge. See Section 2.1.1 of {{HTTP-Authentication}} for discussion.
+- Per-Origin or cross-Origin. Tokens can be constrained to the Origin for
+  which the challenge originated (referred to as per-Origin tokens), or
+  can be used across multiple Origins (referred to as cross-Origin tokens).
+  The set of Origins for which a cross-Origin token is applicable is referred
+  to as the cross-Origin set.
```
cjpatton, line 208: Much clearer, thanks.

draft-ietf-privacypass-architecture.md
```diff
@@ -669,66 +663,39 @@ realizes that a key compromise has occurred then the Issuer should
 generate a new key and make it available to Clients. If possible, it should
 invoke any revocation procedures that may apply for the old key.
 
-## Large Number of Issuers {#servers}
+## Issuer Selection {#servers}
```
cjpatton, line 666: Much better title IMO

cjpatton (approved):

ekinnear (approved):

PR #194, opened 2022-09-16
Clarify issuance protocol input

### LINKED ISSUE #188: Arch: s3.2 client private input
1. The input secrecy requirement suggests that client private input would include the nonce, challenge, but doesn't give any ideas what else might be part of the private input:
  > Client's private input, including the challenge and nonce
2. Why is it important to have a client nonce? Is this something that needs to be spelled out in the architecture document or left to the particular issuance protocol specifications?

chris-wood:
This clarifies the inputs to the issuance protocol as constrained by the current HTTP authentication document. It also notes that these inputs may change in the future to accommodate different issuance protocols and Token formats.

Closes #188 

cc @tfpauly, @nikitaborisov

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -250,7 +250,7 @@ Clients initiate the issuance protocol using the challenge, a randomly
 generated nonce, and public key for the Issuer, all of which are the Client's
 private input to the protocol and ultimately bound to an output Token;
 see Section 2.2. of {{HTTP-Authentication}} for details. Future specifications
-may change or exten the Client's input to the issuance protocol to produce
+may change or extend the Client's input to the issuance protocol to produce
 Tokens with a different structure.
 
 The issuance protocol itself can be any interactive protocol between Client,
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -246,11 +246,16 @@ Each issuance protocol may be different, e.g., in the number and types of
...
-Clients initiate the Token issuance protocol using the challenge, a randomly
-generated nonce, and public key for the Issuer. The Token issuance protocol
-itself can be any interactive protocol between Client, Issuer, or other
-parties that produces a valid authenticator over the Client's input, subject
-to the following security requirements.
+Clients initiate the issuance protocol using the challenge, a randomly
+generated nonce, and public key for the Issuer, all of which are the Client's
+private input to the protocol and ultimately bound to an output Token;
+see Section 2.2. of {{HTTP-Authentication}} for details. Future specifications
+may change or exten the Client's input to the issuance protocol to produce
```
tfpauly, line 253:
```suggestion
may change or extend the Client's input to the issuance protocol to produce
```

PR #195, opened 2022-09-16
Clarify non-collusion requirements

### LINKED ISSUE #190: Arch: non-collusion assumption & threat model
The architecture assumes that any two parties do not collude with each other. I think this is too strong an assumption. In particular, if the Origin does not collude with anyone else, what is the point of unlinkability of the issuance and redemption flows? 

Looking at the use cases, it seems that where the Attester and Issuer are distinct there is a needed non-collusion assumption between them to achieve desirable privacy, but I think the origin colluding with one of the Issuer or Attester should be fine and would be worth considering in scope.

Also, in addition to the non-collusion assumption, are there any other assumptions made on the parties, such as honest-but-curious, or do we consider them to be potentially fully malicious? It would be nice to spell this out.

chris-wood:
This doesn't go overboard and clarify non-collusion assumptions for all models. It simply states the requirement that entities with access to the attestation context and redemption context do not collude, and then generalizes somewhat for simplicity. 

Closes #190.

cc @tfpauly., @nikitaborisov

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -427,7 +427,7 @@ and deployed in a number of ways. This section covers some expected deployment
 models and their corresponding security and privacy considerations.
 
 The discussion below assumes non-collusion between entities that have access to
-the attestation context and redemption context, as collusion between such entities
+the attestation context and entities that have access to the redemption context, as collusion between such entities
 would enable linking of these contexts. Generally, this means that entities operated
 by separate parties do not collude. Mechanisms for enforcing non-collusion are out of
 scope for this architecture.
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -424,7 +424,7 @@ about the Client, including their identity or browsing history. Depending on
 the deployment model, separating these contexts can take different forms. The
 Origin, Attester, and Issuer portrayed in {{fig-overview}} can be instantiated
 and deployed in a number of ways. This section covers some expected deployment
-models and their corresponding security and privacy considerations. 
+models and their corresponding security and privacy considerations.
 
 The discussion below assumes non-collusion between entities that have access to
 the attestation context and redemption context, as collusion between such entities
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -424,10 +424,13 @@ about the Client, including their identity or browsing history. Depending on
...
-discussion below assumes non-collusion between entities when operated by
-separate parties. Mechanisms for enforcing non-collusion are out of scope for
-this architecture.
+models and their corresponding security and privacy considerations. 
+
+The discussion below assumes non-collusion between entities that have access to
+the attestation context and redemption context, as collusion between such entities
+would enable linking of these contexts. Generally, this means that entities operated
+by separate parties do not collude. Mechanisms for enforcing non-collusion are out of
+scope for this architecture.
```
chris-wood, line 433:
```suggestion
models and their corresponding security and privacy considerations.

The discussion below assumes non-collusion between entities that have access to
the attestation context and redemption context, as collusion between such entities
would enable linking of these contexts. Generally, this means that entities operated
by separate parties do not collude. Mechanisms for enforcing non-collusion are out of
scope for this architecture.
```

draft-ietf-privacypass-architecture.md
```diff
@@ -424,10 +424,13 @@ about the Client, including their identity or browsing history. Depending on
...
 Origin, Attester, and Issuer portrayed in {{fig-overview}} can be instantiated
 and deployed in a number of ways. This section covers some expected deployment
-models and their corresponding security and privacy considerations. The
-discussion below assumes non-collusion between entities when operated by
-separate parties. Mechanisms for enforcing non-collusion are out of scope for
-this architecture.
+models and their corresponding security and privacy considerations.
+
+The discussion below assumes non-collusion between entities that have access to
+the attestation context and redemption context, as collusion between such entities
```
tfpauly, line 430:
```suggestion
the attestation context and entities that have access to the redemption context, as collusion between such entities
```

tfpauly (approved):

PR #196, opened 2022-09-16
Add section on timing attacks

### LINKED ISSUE #191: Arch: timing correlation
The privacy considerations section should probably include a discussion on timing correlation. The timing issues are non-trivial here; in particular, in the scenarios where the token is issued online in response to a token challenge (which is suggested in all of the diagrams, and which is required if the origin requires session-bound tokens). In that case, the anonymity set of issuance flows that could match a redemption flow is limited to those that started after the challenge was sent and ended before the token was redeemed. Even if process takes a second (which seems slow to me), you need a fairly popular issuer service to get a non-trivial anonymity set, and any (obviously desirable) improvements in the latency of this process put further pressure on the anonymity set.

I think there should be a caution against deploying PrivacyPass in a setting where the issuance rate can be low, at least in combination with session-bound context.

chris-wood:
Closes #191.

cc @tfpauly, @nikitaborisov

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -728,7 +728,20 @@ hit:
   Output resp
 ~~~
 
-## Centralization
+## Side-Channel Attacks
+
+Side-channel attacks, such as those based on timing correlation, could be
+used to link attestation and redemption contexts together. In particular,
+for interactive tokens that are bound to a Client-specific redemption
+context, the anonymity set of Clients during the issuance protocol consists
+of those Clients that started issuance between the time of the Origin's
+challenge and the corresponding token redemption. Depending on the number
+of Clients using a particular Issuer during that time window, the set can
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -728,7 +728,20 @@ hit:
...
+for interactive tokens that are bound to a Client-specific redemption
+context, the anonymity set of Clients during the issuance protocol consists
+of those Clients that started issuance between the time of the Origin's
+challenge and the corresponding token redemption. Depending on the number
+of Clients using a particular Issuer during that time window, the set can
+be small. Appliations should take such side channels into consideration before
+choosing a particular deployment model and type of token challenge and redemption
+context.
+
+# Centralization
```
tfpauly, line 744: Did you intend to bump this up a level from ## to # ?

tfpauly (approved):

PR #210, opened 2022-10-28
Add quotes around base64 auth params

### LINKED ISSUE #209: Padded base64 needs to quoted as an auth parameter
Julian pointed out the following:

> auth parameters are token or quoted-string. If you use base64-url
> encoding with "=" for padding, the token variant would not work, thus
> the abbreviated examples are extremely misleading (might be good to
> insert complete message examples...).

tfpauly:
Closes #209

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -172,16 +172,22 @@ origin names, they are delimited with commas "," without any whitespace.
 When used in an authentication challenge, the "PrivateToken" scheme uses the
 following attributes:
 
-- "challenge", which is a quoted string containing a base64url-encoded {{!RFC4648}}
-TokenChallenge value. Since the length of the challenge is not fixed, the base64url
-data MUST include padding. This MUST be unique for every 401 HTTP response to prevent
-replay attacks. This attribute is required for all challenges.
+- "challenge", which contains a base64url-encoded {{!RFC4648}} TokenChallenge
+ value. Since the length of the challenge is not fixed, the base64url value MUST
+ include padding. As an Authentication Parameter (`auth-param` from {{!RFC9110}}),
+ the value can be either a token or a quoted-string, and might be required to be a
+ quoted-string if the base64url string includes "=" characters. This challenge value
+ MUST be unique for every 401 HTTP response to prevent replay attacks. This attribute
+ is required for all challenges.
 
-- "token-key", which is a quoted string containing a base64url encoding of the public
-key for use with the issuance protocol indicated by the challenge. Since the length of
-the key is not fixed, the base64url data MUST include padding. This attribute MAY
-be omitted in deployments where clients are able to retrieve the issuer key using
-an out-of-band mechanism.
+- "token-key", which contains a base64url encoding of the public key for
+use with the issuance protocol indicated by the challenge. Since the length of
+the key is not fixed, the base64url value MUST include padding. As an
+Authentication Parameter (`auth-param` from {{!RFC9110}}), the value can be
+either a token or a quoted-string, and might be required to be a
+quoted-string if the base64url string includes "=" characters. This attribute
+MAY be omitted in deployments where clients are able to retrieve the issuer key
+using an out-of-band mechanism.
 
 - "max-age", an optional attribute that consists of the number of seconds for which
 the challenge will be accepted by the origin.
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -172,13 +172,13 @@ origin names, they are delimited with commas "," without any whitespace.
 When used in an authentication challenge, the "PrivateToken" scheme uses the
 following attributes:
 
-- "challenge", which contains a base64url-encoded {{!RFC4648}} TokenChallenge
-value. Since the length of the challenge is not fixed, the base64url data MUST
-include padding. This MUST be unique for every 401 HTTP response to prevent
+- "challenge", which is a quoted string containing a base64url-encoded {{!RFC4648}}
+TokenChallenge value. Since the length of the challenge is not fixed, the base64url
+data MUST include padding. This MUST be unique for every 401 HTTP response to prevent
 replay attacks. This attribute is required for all challenges.
 
-- "token-key", which contains a base64url encoding of the public key for
-use with the issuance protocol indicated by the challenge. Since the length of
+- "token-key", which is a quoted string containing a base64url encoding of the public
+key for use with the issuance protocol indicated by the challenge. Since the length of
 the key is not fixed, the base64url data MUST include padding. This attribute MAY
 be omitted in deployments where clients are able to retrieve the issuer key using
 an out-of-band mechanism.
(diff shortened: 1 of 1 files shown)


chris-wood:
Is this a breaking change? In my reference implementation, I grab everything after the `token=` prefix because that's what the spec said to do.

reschke:
> Is this a breaking change? In my reference implementation, I grab everything after the `token=` prefix because that's what the spec said to do.

Well, it would break implementations that do not use the generic HTTP auth grammar for challenges/credentials.

chris-wood:
> Well, it would break implementations that do not use the generic HTTP auth grammar for challenges/credentials.

So, yes, this is a breaking change then =) Before this change the spec did not specify the grammar precisely.

reschke (commented):
Are there cases where the base64url encoding would not include padding? In these cases quoting is not necessary. The spec should require senders to send conforming messages (thus use quoting when needed), but also require recipients to parse token/quoted-string as defined in the base spec (note that for quoted-string that implies unescaping certain sequences).

reschke:
> > Well, it would break implementations that do not use the generic HTTP auth grammar for challenges/credentials.
> 
> So, yes, this is a breaking change then =) Before this change the spec did not specify the grammar precisely.

Well; it said that it is an HTTP auth scheme, and that, by definition, causes specific syntax requirements.

tfpauly:
Right, if the padding isn't needed, then it would be OK to not quote. Our implementation does handle receiving the quotes or not correctly, so the spec was out of alignment there. This is the right change to make.

chris-wood:
@reschke RFC9110 has this to say about `auth-param` encoding:

> Parameter values can be expressed either as "token" or as "quoted-string" ([Section 5.6](https://greenbytes.de/tech/webdav/rfc9110.html#fields.components)). Authentication scheme definitions need to accept both notations, both for senders and recipients, to allow recipients to use generic parsing components regardless of the authentication scheme.

Most of the implementations of authentication headers I've seen do _not_ use any sort of generic parsing components. Now, I've not seen very many, but enough to conclude that people just reimplement whatever parsing is needed based on the auth type. Do you know how common it is for implementations to rely on these "generic" parsing functions? 

I ask because this change means that servers that were not previously supporting both types of auth-param encodings need to, and there doesn't seem to be any good reason to do so rather than, say, just pick one. If a server is using these generic parsing functions, then they can continue using them. Otherwise, if they're rolling parsing fresh, they have a simpler task at hand.

Assuming clients can also easily pick one option, what's the technical reason for wanting to support both?

reschke:
Apart of not having to rewrite parsers for each scheme?

*Clients* need to use a generic parser, because WWW-A can hold a list of different challenges. To get to the point where a client can pick the "best" challenge, it actually has to consume the full combined field value, and the only way to do that is a parser that is scheme-agnostic.

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -172,13 +172,13 @@ origin names, they are delimited with commas "," without any whitespace.
 When used in an authentication challenge, the "PrivateToken" scheme uses the
 following attributes:
 
-- "challenge", which contains a base64url-encoded {{!RFC4648}} TokenChallenge
-value. Since the length of the challenge is not fixed, the base64url data MUST
-include padding. This MUST be unique for every 401 HTTP response to prevent
+- "challenge", which is a quoted string containing a base64url-encoded {{!RFC4648}}
+TokenChallenge value. Since the length of the challenge is not fixed, the base64url
+data MUST include padding. This MUST be unique for every 401 HTTP response to prevent
 replay attacks. This attribute is required for all challenges.
```
tfpauly, line 178:
```suggestion
- "challenge", which contains a base64url-encoded {{!RFC4648}} TokenChallenge
 value. Since the length of the challenge is not fixed, the base64url value MUST
 include padding. As an Authentication Parameter (`auth-param` from {{!RFC9110}}),
 the value can be either a token or a quoted-string, depending on if the base64url string
 includes "=" characters. This challenge value MUST be unique for every 401 HTTP
 response to prevent replay attacks. This attribute is required for all challenges.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -172,13 +172,13 @@ origin names, they are delimited with commas "," without any whitespace.
...
 
-- "token-key", which contains a base64url encoding of the public key for
-use with the issuance protocol indicated by the challenge. Since the length of
+- "token-key", which is a quoted string containing a base64url encoding of the public
+key for use with the issuance protocol indicated by the challenge. Since the length of
 the key is not fixed, the base64url data MUST include padding. This attribute MAY
 be omitted in deployments where clients are able to retrieve the issuer key using
 an out-of-band mechanism.
```
tfpauly, line 184:
```suggestion
- "token-key", which contains a base64url encoding of the public key for
use with the issuance protocol indicated by the challenge. Since the length of
the key is not fixed, the base64url value MUST include padding. As an
Authentication Parameter (`auth-param` from {{!RFC9110}}), the value can be
either a token or a quoted-string, depending on if the base64url string
includes "=" characters. This attribute MAY be omitted in deployments where
clients are able to retrieve the issuer key using an out-of-band mechanism.
```

draft-ietf-privacypass-auth-scheme.md
reschke: "depending on if the base64url string includes "=" characters" -> "and might be required to ba quoted-string if...."

draft-ietf-privacypass-auth-scheme.md
reschke: (same type of comment here; the recipient needs to handle both formats no matter what)

draft-ietf-privacypass-auth-scheme.md
tfpauly: Updated!

draft-ietf-privacypass-auth-scheme.md
tfpauly: Updated!

tfpauly:
@reschke let me know if this text looks good now

reschke (approved):

chris-wood (commented):
This aligns with RFC9110, but I still don't really like it. In particular, I see no technical reason why we can't be more restrictive here, given that this is a new auth type and, based on my experience, most servers will have to roll new parsing logic for it anyway.

To be concrete, why can't we simply say that clients MUST use auth-params when encoding auth-params to the server? And similarly for parameters from server to client? @reschke, thoughts?

chris-wood (approved):

chris-wood:
Approved and taking my concerns to this [lovely new draft](https://github.com/httpwg/http-extensions/issues/2280)!

reschke:
> To be concrete, why can't we simply say that clients MUST use auth-params when encoding auth-params to the server? And similarly for parameters from server to client? @reschke, thoughts?

Not sure I understand the question. The syntax for the authentication fields is defined in RFC 9110. You really can't change it (while conforming to that spec). All you get to do is naming the scheme and the auth-param names.

chris-wood:
My objection is to the syntax in rFC 9110 itself. But I've settled with addressing that syntax elsewhere.

reschke:
> My objection is to the syntax in RFC 9110 itself.

Then your objection is ~25 years late :-).

chris-wood:
Heh, I'm aware! :)

PR #246, opened 2022-11-30
Challenges may be bound to TLS connections, not HTTP sessions

### LINKED ISSUE #243: Auth: HTTP session
> A TokenChallenge MAY be bound to a specific HTTP session with client

HTTP is a stateless protocol; there is no concept of a 'session'.

tfpauly:
Closes #243

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -348,7 +348,7 @@ authenticator using the public key of the issuer, and validate that the signed
 message matches the concatenation of the client nonce and the hash of a
 valid TokenChallenge. For context-bound tokens, origins store or reconstruct the
 contexts of previous TokenChallenge structures in order to validate the token.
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -348,7 +348,7 @@ authenticator using the public key of the issuer, and validate that the signed
 message matches the concatenation of the client nonce and the hash of a
 valid TokenChallenge. For context-bound tokens, origins store or reconstruct the
 contexts of previous TokenChallenge structures in order to validate the token.
-A TokenChallenge MAY be bound to a specific HTTP session with client, but
+A TokenChallenge MAY be bound to a specific TLS connection from a client, but
```
chris-wood, line 351:
```suggestion
A TokenChallenge MAY be bound to a specific TLS session with a client, but
```
Since it would be reasonable for the challenge to be associated with a ticket, for example.

chris-wood (approved):

PR #248, opened 2022-11-30
Clarify that tokens are generally single-use

### LINKED ISSUE #242: Auth: interaction model
it's not crystal-clear what the interaction model for this auth scheme is.

In many auth schemes, the credential is repeated on each request, to keep the protocol stateless, and to prove the client's identity. 

However, that doesn't seem to be desireable here -- not only would it make all of those responses uncacheable, it would (AIUI) 'spend' a lot of private tokens. OTOH, the 'realm' parameter is explicitly supported (albiet without much detail), which implies the opposite. 

I _think_ the preferred interaction model is that a token is only spent upon an explicit challenge from the server, but it was difficult to discern this from the documentation.

Either way, it'd be good to clarify this.

tfpauly:
Closes #242

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -112,6 +112,13 @@ from a specific issuer ({{challenge}}). Once a client has received a token
 from that issuer, or already has a valid token available, it presents the
 token to the origin ({{redemption}}).
 
+Unlike many authentication schemes in which a client will present the same credentials
+across multiple requests, tokens used with the "PrivateToken" scheme are generally
+single-use credentials, and are not reused. Specific token issuance protocols MAY
+define behavior that allows token reuse, however. In deployment scenarios where origins send
+token challenges to request tokens, origins ought to expect at most one request containing
+a token from the client in reaction to a particular challenge.
+
 ## Token Challenge {#challenge}
 
 Origins send a token challenge to clients in an "WWW-Authenticate" header field with
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -114,10 +114,9 @@ token to the origin ({{redemption}}).
 
 Unlike many authentication schemes in which a client will present the same credentials
 across multiple requests, tokens used with the "PrivateToken" scheme are
-single-use credentials, and are not reused. Specific token issuance protocols MAY
-define behavior that allows token reuse, however. In deployment scenarios where origins send
-token challenges to request tokens, origins ought to expect at most one request containing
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -112,6 +112,13 @@ from a specific issuer ({{challenge}}). Once a client has received a token
 from that issuer, or already has a valid token available, it presents the
 token to the origin ({{redemption}}).
 
+Unlike many authentication schemes in which a client will present the same credentials
+across multiple requests, tokens used with the "PrivateToken" scheme are generally
```
chris-wood, line 116:
```suggestion
across multiple requests, tokens used with the "PrivateToken" scheme are
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -112,6 +112,13 @@ from a specific issuer ({{challenge}}). Once a client has received a token
...
+single-use credentials, and are not reused. Specific token issuance protocols MAY
```
chris-wood, line 117: I think I understand the intent here -- to allow issuance protocols where tokens can be "presented" more than once without privacy regressions -- but this text seems like it would allow clients to literally double spend the same value. That would be a privacy regression. Instead, I would phrase this as something like "spending the same token value more than once allows the origin to link multiple sessions to the same client," and omit mention of what future protocols may do.

chris-wood (changes requested):

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -112,6 +112,13 @@ from a specific issuer ({{challenge}}). Once a client has received a token
 from that issuer, or already has a valid token available, it presents the
 token to the origin ({{redemption}}).
 
+Unlike many authentication schemes in which a client will present the same credentials
+across multiple requests, tokens used with the "PrivateToken" scheme are
+single-use credentials, and are not reused. Specific token issuance protocols MAY
+define behavior that allows token reuse, however. In deployment scenarios where origins send
+token challenges to request tokens, origins ought to expect at most one request containing
+a token from the client in reaction to a particular challenge.
```
tfpauly, line 120:
```suggestion
single-use credentials, and are not reused. Spending the same token value more than once allows the origin to link multiple transactions to the same client. In deployment scenarios
where origins send token challenges to request tokens, origins ought to expect at most one
request containing a token from the client in reaction to a particular challenge.
```

chris-wood (approved):

draft-ietf-privacypass-auth-scheme.md
chris-wood: Looks good!

PR #252, opened 2022-12-02
Clean up auth-scheme usage and references

### LINKED ISSUE #231: Proto: Token Type
table 2 says:

> 

token-type | Integer value of the Token Type, as defined in Section 8.1, as a JSON number
-- | --

and Section 8.1 says:

> This document updates the "Token Type" Registry with the following values.

What is the Token Type registry and what are the constraints on values in the registry? It's implied that values should be integers but JSON numbers are quite flexible, so would it be valid for me to register e.g. a negative integer. If the "Token Type" registry already exists, I'm not able to find it and a link would be useful. If it doesn't exist and this document wishes to create it, then more work is needed to establish that registry and it's procedures.

### LINKED ISSUE #233: Proto: two refernece to Privacy Pass HTTP authentication scheme
There's `[AUTHSCHEME]` and `[HTTP-Authentication]`. I'm guessing these should be collapsed to just one.

chris-wood:
Closes #231
Closes #233

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -184,7 +184,7 @@ Clients provide the following as input to the issuance protocol:
 - Issuer Public Key pkI, with a key identifier `token_key_id` computed as
   described in {{issuer-configuration}}.
 - Challenge value `challenge`, an opaque byte string. For example, this might
-  be provided by the redemption protocol in {{HTTP-Authentication}}.
+  be provided by the redemption protocol in {{AUTHSCHEME}}.
 
 Given this configuration and these inputs, the two messages exchanged in
 this protocol are described below. This section uses notation described in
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -579,7 +576,7 @@ consistency mechanism to help ensure they are not being given unique keys; see
 
 ## Token Type
 
-This document updates the "Token Type" Registry with the following values.
+This document updates the "Token Type" Registry from {{AUTHSCHEME, Section 5.2}} with the following values.
```
LPardue, line 579: You're also missing the `Nid: The length of the token key identifier` that the auth scheme says is required to be populated

chris-wood: Good catch!

chris-wood: Done.

tfpauly (approved):

PR #257, opened 2022-12-02
Add blind RSA variants to the registry

chris-wood:
Builds on #256 -- merge that first.

We might consider using a list to describe the registry contents, rather than a table, now that it has so many entries. @tfpauly, thoughts?


tfpauly (approved):

tfpauly:
I think it's OK as a table still for now, since that matches more what the registry will look like?

chris-wood:
It just renders pretty poorly in the HTML. I think I'll prep a list version as an alternate so we can compare.

PR #259, opened 2022-12-02
Add example of 401 response that supports legacy clients

### LINKED ISSUE #244: Auth: When to use?
If a server wants to use this auth scheme but the client doesn't support it, what's supposed to happen?

Do you have a way for the server to discover that the client does support it?

tfpauly:
Closes #244

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -408,7 +408,7 @@ is a real human user. The origin would send clients a 401 HTTP response that con
 token challenge in a "WWW-Authenticate" header field along with content that contains
 the puzzle to display to the user. Clients that are able to respond with a token will
 be able to automatically return the token and not show the puzzle, while clients that
-either do not support tokens or are unable to fetch tokens at a particular time will
+either do not support tokens or are unable to fetch tokens at a particular time can
 present the user with the puzzle.
 
 To mitigate the risk of deployments becoming dependent on tokens, clients and servers SHOULD grease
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -402,6 +402,15 @@ or issuer is unavailable, or clients might simply not support the requested toke
 Origins SHOULD account for such operational or interoperability failures by offering
 clients an alternative type of challenge such as CAPTCHA for accessing a resource.
 
+For example, consider a scenario in which the client is a web browser, and the origin
+can accept either a token or a solution to a puzzle intended to determine if the client
+is a real human user. The origin would send clients a 401 HTTP response that contains a
+token challenge in a "WWW-Authenticate" header field along with content that contains
```
chris-wood, line 408:
```suggestion
TokenChallenge in a "WWW-Authenticate" header field along with content that contains
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -402,6 +402,15 @@ or issuer is unavailable, or clients might simply not support the requested toke
...
+the puzzle to display to the user. Clients that are able to respond with a token will
+be able to automatically return the token and not show the puzzle, while clients that
+either do not support tokens or are unable to fetch tokens at a particular time will
```
chris-wood, line 411:
```suggestion
either do not support tokens or are unable to fetch tokens at a particular time can
```

chris-wood (approved):

tfpauly:
I'll just merge it and we can revise if needed

PR #284, opened 2022-12-12
Make the media type registrations compact

martinthomson:
Once I worked out how to do this, I did this everywhere I could.  These templates are nuts.


martinthomson:
Oh, and you might want to look https://github.com/ietf-wg-ohai/oblivious-http/pull/230/files regarding "8bit"/"binary".

tfpauly (approved):

PR #291, opened 2022-12-13
Fix RSA references. Make constant lengths explicit.

### LINKED ISSUE #282: Assumptions about length
In the issuance protocol draft, there are a few places where it is *assumed* that a value will be a certain length, but no effort has been made to explicitly set a length.

In [Section 6.1](https://www.ietf.org/archive/id/draft-ietf-privacypass-protocol-07.html#section-6.1):

```
token_input = concat(0x0002, nonce, challenge_digest, token_key_id)
```

Is the 0x0002 two octets?  Sure, you might readily assume that, but nothing says that explicitly.

The same assumption appears in Sections 5.1 and 5.5.

### LINKED ISSUE #283: Broken links to Blind RSA draft
> The rsabssa_blind_sign function is defined in [[BLINDRSA](https://www.ietf.org/archive/id/draft-ietf-privacypass-protocol-07.html#BLINDRSA)], [Section 5.1.2.](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-rsa-blind-signatures-07#section-5.1.2.).

This link is broken in a few ways.  Note that the links has an trailing period in it, which is only part of the problem there; there is no Section 5.1.2 or Section "5.1.2.".  Also, the function is called "BlindSign" in that version of the draft.

chris-wood:
Closes #283
Closes #282

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -226,7 +226,7 @@ with the input challenge and Issuer key identifier as described below:
 ~~~
 nonce = random(32)
 challenge_digest = SHA256(challenge)
-token_input = concat(0x0001, // 2-byte length
+token_input = concat(0x0001, // Token type field is 2 bytes long
                      nonce,
                      challenge_digest,
                      token_key_id)
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -100,7 +100,7 @@ This document additionally uses the terms "Origin" and "Token" as defined in
 {{ARCHITECTURE}}.
 
 Unless otherwise specified, this document encodes protocol messages in TLS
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -221,7 +226,7 @@ with the input challenge and Issuer key identifier as described below:
 ~~~
 nonce = random(32)
 challenge_digest = SHA256(challenge)
-token_input = concat(0x0001,
+token_input = concat(0x0001, // 2-byte length
```
tfpauly, line 229: I find this a bit confusing — it almost sounds like there would be an extra 2-byte length field before this?

How about:

```suggestion
token_input = concat(0x0001, // Token type field is 2 bytes long
```

tfpauly (approved):

PR #294, opened 2022-12-13
Give the directory a proper type

### LINKED ISSUE #271: Table 1 has no anchor in prose
The table appears without the text referring to it, even with an "as follows".  It might be better to reference it directly.

### LINKED ISSUE #274: Media type for the directory
This is a bit of a hard one to decide for me, but I think that you want to define a media type for the JSON directory.

I know that in most cases the content will be served from a .well-known location, but that isn't always.

### LINKED ISSUE #279: xxx well-known URI
The caption for [Table 3](https://www.ietf.org/archive/id/draft-ietf-privacypass-protocol-07.html#table-3) is: "'xxx' Well-Known URI", which is probably wrong.

chris-wood:
Closes #279
Closes #274
Closes #271

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -740,6 +740,7 @@ Author:
 Change controller:
 
 : IESG
+{: spacing="compact"}
 
 ### "application/private-token-request" media type
 
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -621,7 +621,7 @@ following values.
 | URI Suffix  | Change Controller  | Reference | Status | Related information |
 |:------------|:-------------------|:----------|:-------|:--------------------|
 | token-issuer-directory | IETF | [this document] | permanent | None |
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -658,11 +664,83 @@ This document updates the "Token Type" Registry from
...
+
+: N/A
+
+Fragment identifier considerations:
+
+: N/A
+
+Additional information:
+
+: <dl>
```
tfpauly, line 717:
```suggestion
: <dl spacing="compact">
```

draft-ietf-privacypass-protocol.md
```diff
@@ -658,11 +664,83 @@ This document updates the "Token Type" Registry from
...
+
+: N/A
+
+Author:
+
+: see Authors' Addresses section
+
+Change controller:
+
+: IESG
```
tfpauly, line 742:
```suggestion
: IESG
{: spacing="compact"}
```

tfpauly (approved):

PR #296, opened 2022-12-13
Just make it a URL

### LINKED ISSUE #288: Issuer Request URI: URL? "text/plain"?
> Issuer Request URI: A token request URL for generating access tokens.  For example, an Issuer URL might be https://issuer.example.net/example-token-request.  This parameter uses resource media type "text/plain".

Since this URI is intended to be used as a URL later in the protocol -- and the description in fact calls it a URL -- would it be cleaner to name the parameter "Issuer Request URL"/`issuer-request-url`?

What "thing" is actually `text/plain`? The directory itself is `application/json`, the resulting content is JSON (whose strings do not carry a notion of media types), and the token request/response using this URI use `application/private-token-request`/`application/private-token-response`.

### LINKED ISSUE #300: Issuer directories are underspecified
There are several related issues in https://github.com/ietf-wg-privacypass/base-drafts/blob/main/draft-ietf-privacypass-protocol.md#configuration-setup:

1. Should the well-known location `/.well-known/token-issuer-directory` require mTLS? The two well-known public implementations previously highlighted by Apple disagree on this point.

Our opinion: mTLS should not be required for the token issuer directory.

2. `issuer-request-uri` is underspecified. Should this be an absolute URI, a relative URI, or allow either format?

Our opinion: We prefer relative URIs, and think either option should be supported. A relative URI is also used by at least one well-known public implementation.

chris-wood:
Closes #288
Closes #300


tfpauly:
Are the changes to other files on this one intentional @chris-wood ?

chris-wood:
Oh, hrm, no, they're not. I'll remove them.

chris-wood:
@tfpauly fixed!

tfpauly (approved):

PR #297, opened 2022-12-13
Address caching and race conditions

### LINKED ISSUE #287: Safe addition/removal of token_types by an issuer
I believe there is a race condition if an issuer tries to add or remove `token_type`s in its directory.

When adding a token type, a client may see the a key for the new `token_type` in the directory and use it even though the origin may have a slightly stale directory.

When removing a token, a client may see still see the key for the old `token_type` in a slightly stale directory and use it even the the origin may already have refreshed its directory.

One potential way to address this would be to extend or replace key ordering with a designation of which key(s) are "active." The client would only pick "active" keys to make requests to an issuer, but origins would verify against all listed keys. A `token_type` could then be "rotated" in/out of use by having it appear for some time as inactive before activating it or removing it entirely.

### LINKED ISSUE #289: caching of issuer directories
Most likely, clients and origins will be caching issuer directories.

Is it worth offering any guidance to client/origin implementations regarding caching? Is honoring cache-control response headers sufficient? And if so, should that be called out explicitly as a reminder to implementers (who might be building a custom cache or perhaps caching the parsed JSON content separately from an existing HTTP layer)?

Is it worth offering any guidance to issuers about how long they should assume directory changes will take to reach ~100% of clients and issuers? (My gut says no to this one -- but I am throwing it out in case others think it is worthwhile.)

chris-wood:
Closes #287 
Closes #289


draft-ietf-privacypass-protocol.md
```diff
@@ -158,6 +158,25 @@ Issuer directory resources have the media type "application/json"
...
+
+~~~
+  Cache-Control: max-age=86400
+~~~
+
+Consumers of the Issuer directory resource SHOULD follow the usual HTTP caching
+{{!RFC9111}} semantics when processing this resource. Long cache lifetimes may
+result in use of stale Issuer configuration information, whereas short
+lifetimes may result in decreased performance. When use of an Issuer
+configuration results to token issuance failures, e.g., because the
```
chris-wood, line 176:
```suggestion
configuration results in token issuance failures, e.g., because the
```

draft-ietf-privacypass-protocol.md
```diff
@@ -158,6 +158,25 @@ Issuer directory resources have the media type "application/json"
...
+
+~~~
+  Cache-Control: max-age=86400
+~~~
+
+Consumers of the Issuer directory resource SHOULD follow the usual HTTP caching
+{{!RFC9111}} semantics when processing this resource. Long cache lifetimes may
+result in use of stale Issuer configuration information, whereas short
+lifetimes may result in decreased performance. When use of an Issuer
+configuration results in token issuance failures, e.g., because the
```
tfpauly, line 176: Who is detecting a failure due to stale config? The origin or the client?

chris-wood: Either of them could presumably detect a failure, though in practice it's usually the client.

tfpauly: I think it might be useful to clarify this; since the client also gets the key information from a challenge, it seems a bit odd that the caching would mess it up?

chris-wood: Well if the origin has a stale cached copy of the key and gives it to the client in the challenge, the client's issuance request would fail. I suppose we can be more specific, but it just seems like more words than required here.

draft-ietf-privacypass-protocol.md
```diff
@@ -200,6 +200,32 @@ Issuer directory resources have the media type
...
+configuration information is too stale, the directory SHOULD be fetched and
+revalidated.
+
+# Token Challenge Requirements
+
+Clients receive challenges for tokens, as described in
+{{!AUTHSCHEME=I-D.ietf-privacypass-auth-scheme}}. The basic token issuance
+protocols described in this document can be interactive or non-interactive,
+and per-origin or cross-origin.
```
chris-wood, line 227:
```suggestion
```

Smithad499 (approved):

PR #301, opened 2022-12-14
Improve privacy and consistency discussion

### LINKED ISSUE #278: Partitioning of users
There are claims in [Section 7 of the protocol](https://www.ietf.org/archive/id/draft-ietf-privacypass-protocol-07.html#section-7) about the value of truncating the issuer's key identifier that I don't think hold outside of having some sort of consistency.

Let's say that I - your hypothetical attacker - want to partition users into 257 groups.  I give the first 255 groups a unique key for which the last byte of `token_key_id = SHA256(concat(0x0001, SerializeElement(pkI)))` is distinct.  For the last two users, I can generate two keys that map to the remaining byte value.

For the private validation setting, I can issue clients with unique values for `pkI` and tokens that match.  When validating the token, that's me (the attacker) again, so I can validate against both keys.  The complexity of my operation has doubled, but it's a linear search and maybe I can keep the numbers down to manage overall costs.

In a public validation setting, I have to ensure that any entity that might validate the tokens also gets a matching value for `pkI`, or that they are complicit in the attack and so are willing to do the extra work.  Generally, this is only the client still, so this is probably achievable, though there is a risk of an independent audit catching this.

Going from 8 bits to 9 bits of partition space doubles the work for any validator, and each additional bit doubles again.  That might be OK, but the claims would appear to be somewhat more absolute than this result implies:

> Since the key ID is truncated to a single byte, an Issuer can partition the set of Clients into at most 256 different anonymity sets.

Also, even this isn't a great outcome either.  256 is pretty large, and the partitioning allocation (again, absent consistency) is under attacker control.

### LINKED ISSUE #280: Redemption
The protocol descriptions in the protocol draft (Sections 5 and 6) don't include a heading for token redemption.  It looks like the finalization step is where some of that happens, but it seems fairly crucial for the draft to explain that clients to remove blinding and then construct a token that can be redeemed...*and then they send it to the origin*.

There is a similar problem in the architecture draft where the "redemption" protocol doesn't show the challenge-response flow, or even a spontaneous flow.  That is left for the section following, which covers issuance.

chris-wood:
Closes #278 
Closes #280

The new aligned diagrams that match across the architecture, protocol, and auth-scheme documents should address #280 by making it clear how the tokens are used.

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -46,6 +46,14 @@ informative:
     author:
       ins: N. Sullivan
       org: Cloudflare
+  Dingledine2004:
+    title: "Tor: The Second-Generation Onion Router"
+    date: 2004-08
+    target: "https://svn.torproject.org/svn/projects/design-paper/tor-design.html"
+    author:
+      - ins: R. Dingledine
+      - ins: N. Mathewson
+      - ins: P. Syverson
   HIJK21:
     title: "PrivateStats: De-Identified Authenticated Logging at Scale"
     target: https://research.fb.com/privatestats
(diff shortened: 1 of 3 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -282,11 +296,12 @@ See {{extensions}} for requirements on new issuance protocol variants and
 related extensions.
 
 Clients obtain the Issuer public key directly from the Origin using the process
-described in {{AUTHSCHEME}}. Clients MAY apply some form of key
+described in {{AUTHSCHEME}}. Clients MAY apply some form of
```
tfpauly, line 299: Should this be elevated to a SHOULD instead of a MAY?

chris-wood: I think it depends on your threat model. Maybe we just drop normative language here altogether and say "Clients can apply some form of consistency check... Failure to do so may allow Client-specific keys to impact the Client anonymity set."

draft-ietf-privacypass-architecture.md
```diff
@@ -619,16 +610,24 @@ only a single Origin, then per-Origin tokens are not appropriate.
...
+context. This separation means that the attestation context (or redemption
+context) for a Client does not uniquely identify the redemption context (or
+attestation) context. In other words, for a given attestation context, there
+exists a sufficiently large anonymity set of Clients that share the same
+redemption context, and vice versa.
+
+Maintaining these anonymity sets is a fundamental requirement for Privacy Pass
+deployments. In practice, there are a number of ways in which these anonymity
+sets can be partitioned. For example, as discussed in {{deployment}}, failure
+to use a privacy-enhancing proxy system such as Tor when interacting with
```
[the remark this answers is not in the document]
chris-wood: We don't currently include a citation for Tor when used elsewhere, but we can drop one in.

draft-ietf-privacypass-architecture.md
chris-wood: I made this change -- how's it look?

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
chris-wood: Added a citation.

draft-ietf-privacypass-architecture.md
tfpauly: Looks good!

tfpauly (approved):

dvorak42 (approved):

PR #310, opened 2023-01-09
Clarify how architecture is introduced

### LINKED ISSUE #304: Stop leaning so hard on the authscheme draft
The architecture references \[AUTHSCHEME\] for key concepts far more than I expected.  For instance:

> As described in \[AUTHSCHEME\], there are a number of ways in which the token may vary, including:

Or:

> The Privacy Pass issuance protocol is a two-message protocol that takes as input a TokenChallenge from the redemption protocol ([AUTHSCHEME](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#AUTHSCHEME)], [Section 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-privacypass-auth-scheme-07#section-2.1)) and produces a Token ([[AUTHSCHEME](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#AUTHSCHEME)], [Section 2.2](https://datatracker.ietf.org/doc/html/draft-ietf-privacypass-auth-scheme-07#section-2.2)),

Can the architecture define the messages in their own right, then point to \[AUTHSCHEME\] for example implementations?

### LINKED ISSUE #305: Diagrams might be unclear
In this image:
image

It isn't very clear where messages originate and who gets to see them.  In particular, it isn't clear where the TokenRequest originates (the Client) and whether the Attester gets to see it (¯\\(°\_o)/¯).

tfpauly:
Closes #304
Closes #305


draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,69 @@ The following terms are used throughout this document.
   attested to by the Attester.
 - Attester: An entity that attests to properties of Client for the
   purposes of token issuance.
+- Redemption context: The interactions and set of information shared
+between the Client and Origin.
+- Issuance context: The interactions and set of information shared
+between the Client, Attester, and Issuer.
```
chris-wood, line 138: This is good! Should we also add the attestation context, or were you thinking that the issuance context would subsume the attestation context?

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,69 @@ The following terms are used throughout this document.
...
+
+The typical interaction flow for Privacy Pass tokens uses the
+following steps:
+
+1. A Client interacts with an Origin by sending an HTTP request.
+The Origin sends an HTTP response that contains a token challenge
+that indicates a specific Issuer to use.
+Note that the request might be made as part of accessing a
+resource normally, or with the specific intent of triggering a token
+challenge.
```
chris-wood, line 156:
```suggestion
resource normally, or with the specific intent of triggering a Token
Challenge.
```

If we're introducing Token Request/Response here (to be defined in the issuance doc), do we also want to introduce Token Challenge and Token?

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,69 @@ The following terms are used throughout this document.
...
+
+4. The Issuer generates a Token Response based on the Token Request, which
+is returned to the Client. Upon receiving the Token Response, the Client
+generates a token by cryptographically transforming the content of the
+response so that the token can be validated using a per-Issuer key, but
+cannot be linked to the content of the Token Response.
+
+5. If the Client has a token, it includes it in a subsequent HTTP
+request to the Origin, as authorization. This token is sent only once.
+The Origin validates that the token was generated by the expected Issuer.
```
chris-wood, line 178:
```suggestion
challenge, it can skip to step 5 and redeem its token. Otherwise, it
invokes the issuance protocol to request a token from the designated Issuer.

3. The first step in the issuance protocol is attestation. Specifically, the Attester
performs attestation checks on the Client. These checks
could be proof of solving a CAPTCHA, device trust, hardware attestation,
etc (see {{attester}}). 

4. If attestation succeeds, the client creates a Token Request to send
to the designated Issuer (generally via the Attester). The Attester and Issuer
might be functions on the same server, depending on the deployment model
(see {{deployment}}). Depending on the details of Attestation, the Client can
send the Token Request to the Attester alongside any attestation information.
If attestation fails, the Client receives an error and issuance aborts without a token.

4. The Issuer generates a Token Response based on the Token Request, which
is returned to the Client (generally via the Attester). Upon receiving the Token
Response, the Client computes a token from the token challenge and Token
Response. This token can be validated by anyone with the per-Issuer key, but
cannot be linked to the content of the Token Request or Token Response.

5. If the Client has a token, it includes it in a subsequent HTTP
request to the Origin, as authorization. This token is sent only once.
The Origin validates that the token was generated by the expected Issuer
and has not already been redeemed for the corresponding token challenge.
If the Client does not have a token, perhaps because issuance failed, the client
does not reply to the Origin's challenge with a new request.
```

I made a couple of changes in this big suggestion, with perhaps the biggest one being around the attestation language. The current text seemed to imply that attestation was a non-interactive thing that has to be done alongside sending the Token Request, but I don't think that's true. It seems plausible for these to run in sequence, e.g., for the client and attester to first do attestation, and then if that succeeds, for the client and attester to exchange a token request and token response.  

I also added some language in step 5 about what the client does in the event of issuance failure.

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,69 @@ The following terms are used throughout this document.
...
+
+~~~ aasvg
+    Origin           Client      Attester     Issuer
+                   +---------------------------------.
+     Request <-----+                                  |
+                   |                                  |
+ TokenChallenge --->                                  |
+                   | TokenRequest ---------->         |
+                   |             (attest)             |
```
chris-wood, line 187:
```suggestion
                   | <--- (attest) --->             |
                   | TokenRequest ---------->         |
```

I think it's important to make clear that attestation can happen before the token request and response, and not necessarily something in parallel. 

Also, we use different notation to show the process of attestation in the deployment diagrams ("----> Attest"). I'm wondering if there's a better way to draw this to make it clear that attestation is _not_ specified here, is an entirely separate and deployment specific protocol, etc. Maybe something like this?

~~~ aasvg
    Origin           Client      Attester     Issuer
     Request <-----+                                  
                                                       
 TokenChallenge --->                                  
                    +-----------------.
                    |    Attestation   |
                     `----------------'  
                      TokenRequest ---------->
~~~

chris-wood (commented):
I like the new flow. This covers the basic interaction models pretty well! The one thing I think we need to improve (visually) is how we show the attestation step during issuance. I offered a suggestion but it's not the best.

draft-ietf-privacypass-architecture.md
tfpauly: I did view the issuance context as being a superset of the attestation context, but if you think it should be here too, that's fine!

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,76 @@ The following terms are used throughout this document.
...
+challenge.
+
+2. If the Client already has a token available that satisfies the token
+challenge, it can skip to step 5 and redeem its token. Otherwise, it
+invokes the issuance protocol to request a token from the designated Issuer.
+
+3. The first step in the issuance protocol is attestation. Specifically, the Attester
+performs attestation checks on the Client. These checks
+could be proof of solving a CAPTCHA, device trust, hardware attestation,
+etc (see {{attester}}).
```
tfpauly, line 165:
```suggestion
etc (see {{attester}}).
```

draft-ietf-privacypass-architecture.md
tfpauly: How about this as the full picture?
```
    Origin           Client         Attester     Issuer
    
     Request <------                                  
                                                       
 TokenChallenge --->   
                      <- (Attestation) ->                               
  
                      TokenRequest -------------->
                              <-------------- TokenResponse
 Request + Token <--
```

elreydetodo: This diagram has been confusing me since the very first time I looked at the draft. Since the process initiates with a client, can we please put the client on the leftmost column?

tfpauly: The issue is that the client has an interaction with the origin on one hand, and with the attester/issuer on the other, and they are separated contexts.

@chris-wood thoughts?

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,76 @@ The following terms are used throughout this document.
...
+                   | <--- (attest) --->             |
+                   | TokenRequest ---------->         |
+                   |     <------------  TokenResponse |
+ Request + Token <-+                                  |
+                    `--------------------------------'
+                   -----------------+-------------------
+ -----------+-------------          |
+            |                       |
+        Redemption               Issuance
+         Protocol                Protocol
```
tfpauly, line 202:
```suggestion
    Origin           Client         Attester     Issuer
    
     Request <------                                  
                                                       
 TokenChallenge --->   
                      <- (Attestation) ->                               
  
                      TokenRequest -------------->
                              <-------------- TokenResponse
 Request + Token <--
```

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,71 @@ The following terms are used throughout this document.
...
+    Origin           Client         Attester     Issuer
+    
+     Request <------                                  
+                                                       
+ TokenChallenge --->   
+                      <- (Attestation) ->                               
+  
+                      TokenRequest -------------->
+                              <-------------- TokenResponse
+ Request + Token <--
```
tfpauly, line 197:
```suggestion
    Origin           Client         Attester     Issuer

     Request <------

 TokenChallenge --->
                      <- (Attestation) ->
  
                      TokenRequest -------------->
                              <-------------- TokenResponse
 Request + Token <--
```

draft-ietf-privacypass-architecture.md
tfpauly:
```suggestion
between the Client, Attester, and Issuer.
- Attestation context: The interactions and set of information shared between
the Client and Attester only, for the purposes of attesting the vailidity of the Client.
```

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,73 @@ The following terms are used throughout this document.
...
+does not reply to the Origin's challenge with a new request.
+
+~~~ aasvg
+    Origin           Client         Attester     Issuer
+
+     Request <------
+
+ TokenChallenge --->
+                      <- (Attestation) ->
+
```
tfpauly, line 196:
```suggestion

```

draft-ietf-privacypass-architecture.md
```diff
@@ -174,14 +132,72 @@ The following terms are used throughout this document.
...
+Response, the Client computes a token from the token challenge and Token
+Response. This token can be validated by anyone with the per-Issuer key, but
+cannot be linked to the content of the Token Request or Token Response.
+
+5. If the Client has a token, it includes it in a subsequent HTTP
+request to the Origin, as authorization. This token is sent only once.
+The Origin validates that the token was generated by the expected Issuer
+and has not already been redeemed for the corresponding token challenge.
+If the Client does not have a token, perhaps because issuance failed, the client
+does not reply to the Origin's challenge with a new request.
```
tfpauly, line 187:
```suggestion
2. If the Client already has a token available that satisfies the token
challenge, it can skip to step 6 and redeem its token. Otherwise, it
invokes the issuance protocol to request a token from the designated Issuer.

3. The first step in the issuance protocol is attestation. Specifically, the Attester
performs attestation checks on the Client. These checks
could be proof of solving a CAPTCHA, device trust, hardware attestation,
etc (see {{attester}}).

4. If attestation succeeds, the client creates a Token Request to send
to the designated Issuer (generally via the Attester). The Attester and Issuer
might be functions on the same server, depending on the deployment model
(see {{deployment}}). Depending on the details of Attestation, the Client can
send the Token Request to the Attester alongside any attestation information.
If attestation fails, the Client receives an error and issuance aborts without a token.

5. The Issuer generates a Token Response based on the Token Request, which
is returned to the Client (generally via the Attester). Upon receiving the Token
Response, the Client computes a token from the token challenge and Token
Response. This token can be validated by anyone with the per-Issuer key, but
cannot be linked to the content of the Token Request or Token Response.

6. If the Client has a token, it includes it in a subsequent HTTP
request to the Origin, as authorization. This token is sent only once.
The Origin validates that the token was generated by the expected Issuer
and has not already been redeemed for the corresponding token challenge.
If the Client does not have a token, perhaps because issuance failed, the client
does not reply to the Origin's challenge with a new request.
```

draft-ietf-privacypass-architecture.md
tfpauly: We do in the diagram, but here I was using it in a bit more colloquial sense.

chris-wood (approved):

PR #312, opened 2023-01-25
Promote and reframe privacy properties

### LINKED ISSUE #306: Metadata is a privacy risk too
[Section 3.2.3](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-3.2.3) doesn't really acknowledge this.  And [this](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-3.2.3-4) is a bit of a warning flag:

> Metadata may also be arbitrarily long or bounded in length. The amount of permitted metadata may be determined by application or by the underlying cryptographic protocol.

A forward reference to [Section 5.1](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-5.1) might help.

### LINKED ISSUE #307: Adversarial control over anonymity set apportionment
There is an attack on some of the deployment models that isn't well acknowledged in [Section 5.1](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-5.1) or [Section 5.2](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-5.2).  In this attack, there are $n$ bits of metadata (where $n$ isn't necessarily ordinal) or $2^n$ different keys.

In this model, if the Issuer is able to identify users, they can identify $2^n-1$ users for special treatment at the Origin.  Those users get assigned a unique value for metadata (or use a key that is unique to them), so that they can be reidentified at the Origin.  The remaining identifier is assigned to all other users.

I was initially only a little concerned about this style of attack.  This only applies to those deployment models where the Issuer is exposed to information about Clients.  And the proposed 6 metadata values only permit 5 users being linked this way.  That seemed fine.  If there is no way in which a value used for targeting cannot be reused at a high rate.

...

There is an interactive flow (see the [Private State Token](https://wicg.github.io/trust-token-api/#browser-issue-steps) issuance request), so I believe that those $2^n-1$ values could be reused for tracking on a relatively short time scale.

The Issuer and Origin can coordinate their efforts so that when one of the targeted values is used, it is redeemed nearly immediately, then released back for reuse.

The way this works is that on each round the Issuer picks $2^n-1$ issuance requests and allocates one of the special metadata values to those users.  The issuer can now reidentify the user that it issued these tokens to.  The Issuer provides the Origin with an endpoint that says "I just got token with metadata X, what user did you allocate it to?" for which querying has the side effect of releasing metadata X for reuse.  This provides the Issuer a way to provide the Origin with the identity - as known to the Issuer - of those $2^n-1$ users.

This overall process can operate continuously with a flow of Issuer->Client->Origin->Issuer.  Consequently, the number of users that can be linked with this approach is very high as the time where each targeting value is outstanding is low (this amounts to a single round trip to the user).  Stalls would occur if clients fail to complete their part in this, but that should be rare.  You might get something like $(2^n-1)/RTT$ users per second linked, depending on the token issuance rate.  Issuers with millions of users will find it takes a while to get through them all, but with 6 values, that's 2 million a day or thereabouts if you assume a modest 250ms RTT.

I'm not sure how to defend against this.  Including limited metadata is a direct goal, but this attack operates entirely within the envelope that allowing metadata creates.  As this is a direct goal, consistency can't help.

If metadata were removed and diverse keys could not be used for metadata -- that is, keys are strictly sequenced and the consistency mechanism ensured that the Issuer couldn't switch back and forth, perhaps by marking key cutovers with a firm time of effect -- then maybe this is more tractable.  There would be a brief period at each key changeover where maybe some number of users might be targeted by exploiting clock skew, but that's all.  Losing metadata seems pretty bad though.

### LINKED ISSUE #308: Bounding Issuer count
Should [this](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-5.3-3) be "for each Origin"?

> Clients SHOULD bound the number of Issuers they are willing to request tokens from at any given time. The exact bound depends on the deployment model and number of Clients, i.e., having a very large Client base could potentially allow for larger values. Issuer replacements should only occur with the same frequency as config rotations as they can lead to similar losses in anonymity if clients still hold redemption tokens for previously active Issuers.

The previous paragraph says "For example, these 4 Issuers could be different for each Client."  But doesn't really explain how that might be exploited in an attack on Client privacy.

I think that there is an unstated assumption here: the Client has interactions with multiple Origins and uses a single store of tokens from a limited set of Issuers.  This is not at all clear.  Much of the text of the document concentrates on the challenge-based flow, where an Origin asks the Client to fetch a token from the Issuer.   In that context, this text makes little sense.

It's easy to skip over [this](https://ietf-wg-privacypass.github.io/base-drafts/draft-ietf-privacypass-architecture.html#section-5.3-1):

> If Clients obtain tokens from many Issuers, and Origins later challenge a Client for a token from each Issuer, Origins can learn information about the Client.

That ordering is important.  If you allow there to be a shared store of tokens and Origins can ask "do you have a token from Issuer X?", then this creates shared state that can be queried.  That is a far more dangerous baseline.  For that, there needs to be a global cap on the number of Issuers, or there is some information being released.

On the other hand, I'm not sure that we need the ability to check whether tokens are present.  I think that it is sufficient for the Origin to request a token.  There is value in having a token cache, but caches can operate without necessarily leaking that they are in use.  That doesn't totally eliminate time as a factor, but it might help.

In any case, the document could do a better job of laying out the various expected interaction models up front.  This cached mode is very poorly explained and something of a surprise.

chris-wood:
Closes #306
Closes #307 
Closes #308

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -142,13 +142,12 @@ the Client and Attester only, for the purposes of attesting the vailidity of the
 # Architecture
 
 The Privacy Pass architecture consists of four logical entities --
-Client, Origin, Issuer, and Attester -- that work in concert as
+Client, Origin, Issuer, and Attester -- that work in concert
 for token issuance and redemption. This section describes the purpose
 of token issuance and redemption and the requirements on the relevant
 participants.
 
-The typical interaction flow for Privacy Pass tokens uses the
-following steps:
+The typical interaction flow for Privacy Pass uses the following steps:
 
 1. A Client interacts with an Origin by sending an HTTP request.
 The Origin sends an HTTP response that contains a token challenge
@@ -158,8 +157,9 @@ resource normally, or with the specific intent of triggering a token
 challenge.
 
 2. If the Client already has a token available that satisfies the token
-challenge, it can skip to step 6 and redeem its token. Otherwise, it
-invokes the issuance protocol to request a token from the designated Issuer.
+challenge, e.g., because the Client has a cache of previously issued tokens,
+it can skip to step 6 and redeem its token. Otherwise, it invokes the issuance
+protocol to request a token from the designated Issuer.
 
 3. The first step in the issuance protocol is attestation. Specifically, the Attester
 performs attestation checks on the Client. These checks
@@ -198,6 +198,42 @@ does not reply to the Origin's challenge with a new request.
  Request + Token <--
 ~~~
 {: #fig-overview title="Privacy pass redemption and issuance protocol interaction"}
+
+The end-to-end flow for Privacy Pass involves three different types of contexts:
+
+- Redemption context: The interactions and set of information shared
+between the Client and Origin.
+- Issuance context: The interactions and set of information shared
+between the Client, Attester, and Issuer.
+- Attestation context: The interactions and set of information shared between
+the Client and Attester only, for the purposes of attesting the vailidity of the Client.
+
+The privacy goals of Privacy Pass are oriented around unlinkability based on these
+contexts. In particular, Privacy Pass aims to achieve Client unlinkability. This
+means that given two redemption (or issuer) contexts, the Origin (or Issuer) cannot
+determine if both redemption (or issuance) contexts correspond to the same Client
+or two different Clients. Informally, this means that a Client in a redemption context
+is indistinguishable from any other Client that might use the same redemption context.
+The set of Clients that share the same redemption (or issuance) context is referred
+to as an anonymity set.
+Depending on the deployment model, Privacy Pass might also aim to achieve Origin
+unlinkability. Similar to Client unlinkability, this means that given two attestation
+contexts, the Attester cannot determine if both contexcts correspond to the same
+Origin or two different Origins. The set of Clients that share the same attestation
+context is referred to as an anonymity set.
+
+At a high level, these properties ensure that no single party amongst the Attester,
+Issuer, or Origin can link client identifying information to client activity, e.g.,
+the origin being accessed.
+
+The manner in which Client and Origin unlinkability are achieved depends on the deployment
+model, type of attestation, and issuance protocol details. For example, as discussed
+in {{deployment}}, failure to use a privacy-enhancing proxy system such as Tor when
+interacting with Attesters, Isuers, or Origins allows the set of possible Clients to be
+partitioned by the Client's IP address, and can therefore lead to Client unlinkability
+violations. Similarly, malicious Origins may attempt to link two redemption contexts
+together by using Client-specific Issuer public keys. See {{deployment}} and
+{{privacy}} for more information.
 
 ## Redemption Protocol
 
@@ -274,9 +310,8 @@ as input a TokenChallenge from the redemption protocol
 ~~~ aasvg
    Origin              Client      Attester     Issuer
                    +-----------------------------------.
-TokenChallenge --->| Attest -------->                   |
+TokenChallenge --->| <- (Attestation) ->                |
                    | TokenRequest ---------------->     |
-                   |                         (evaluate) |
      Token    <----+     <--------------- TokenResponse |
                     `----------------------------------'
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -199,6 +199,42 @@ does not reply to the Origin's challenge with a new request.
...
 {: #fig-overview title="Privacy pass redemption and issuance protocol interaction"}
 
+The end-to-end flow for Privacy Pass involves three different types of contexts:
+
+- Redemption context: The interactions and set of information shared
+between the Client and Origin.
+- Issuance context: The interactions and set of information shared
+between the Client, Attester, and Issuer.
+- Attestation context: The interactions and set of information shared between
+the Client and Attester only, for the purposes of attesting the vailidity of the Client.
```
tfpauly, line 209: Hm, this seems to be duplicated from the terminology above. Do we want to remove them there?

draft-ietf-privacypass-architecture.md
```diff
@@ -199,6 +199,42 @@ does not reply to the Origin's challenge with a new request.
...
+
+The privacy goals of Privacy Pass are oriented around unlinkability based on these
+contexts. In particular, Privacy Pass aims to achieve Client unlinkability. This
+means that given two redemption (or issuer) contexts, the Origin (or Issuer) cannot
+determine if both redemption (or issuance) contexts correspond to the same Client
```
tfpauly, line 214: The redemption/issuer split here is a bit confusing to read. I know it would be redundant to explain both, but it may be easier to read.

draft-ietf-privacypass-architecture.md
chris-wood: I considered it, yeah.

draft-ietf-privacypass-architecture.md
chris-wood: ACK -- I'll try to unwrap that in a followup commit.

chris-wood: OK, I split unlinkability into three distinct properties: Origin-Client, Issuer-Client, and Attester-Origin unlinkability. How does this read now?

draft-ietf-privacypass-architecture.md
```diff
@@ -199,6 +193,44 @@ does not reply to the Origin's challenge with a new request.
...
+Origin or two different Origins. The set of Clients that share the same attestation
+context is referred to as an anonymity set.
+
+At a high level, these properties ensure that no single party amongst the Attester,
+Issuer, or Origin can link client identifying information to client activity, e.g.,
+the origin being accessed.
+
+The manner in which Origin-Client, Issuer-Client, and Attester-Origin unlinkability are achieved depends on the deployment
+model, type of attestation, and issuance protocol details. For example, as discussed
+in {{deployment}}, failure to use a privacy-enhancing proxy system such as Tor when
```
tfpauly, line 227: I think this is now the earliest mention of Tor in the document? If so, we should move the reference here.

draft-ietf-privacypass-architecture.md
```diff
@@ -683,76 +717,47 @@ must balance this against the reduction in Client privacy. In general,
 bounding the metadata permitted ensures that it cannot uniquely identify
```
tfpauly, line 717: I wonder if we should call out here or in the next section that many token types won't have any metadata, and that we have an implicit recommendation to avoid metadata unless necessary and justified and properly analyzed...

draft-ietf-privacypass-architecture.md
```diff
@@ -683,76 +717,47 @@ must balance this against the reduction in Client privacy. In general,
...
-## Side-Channel Attacks
+the same configuration. Issuer configuration updates, e.g., due to key rotation,
+are an important part of hedging against long-term private key compromise.
+In general, key rotations represent a trade-off between Client
+privacy and Issuer security. Therefore, it is important that key rotations occur
+on a regular cycle to reduce the harm of an Issuer key compromise.
+
+Lastly, if there are a large number of selected or trusted Issuers,
+and Origins accept all of them, segregation can occur. If Clients obtain tokens
+from many Issuers, and Origins later challenge a Client for a token from each
```
tfpauly, line 743: Can we also make this clear that it is conditional on the Client being willing to send tokens from all of those issuers? (Since it can just choose not to use those tokens...)

draft-ietf-privacypass-architecture.md
```diff
@@ -683,76 +717,47 @@ must balance this against the reduction in Client privacy. In general,
...
+that a Client holds essentially corresponds to a bit of information about the
+Client that Origin learns. Therefore, there is an exponential loss in anonymity
+relative to the number of Issuers.
+
+The fundamental problem here is that the number of possible issuance configurations,
+including the keys in use and the Issuer identities themselves, can partition the Client
+anonymity set. To mitigate this problem, Clients SHOULD bound the number of active
+issuance configurations. Moreover, Clients SHOULD employ some form of consistency
+mechanism to ensure that they receive the same configuration information and are
+not being actively partitioned into smaller anonymity sets. See {{CONSISTENCY}} for
```
tfpauly, line 754: I guess my comment above could turn into another recommendation here to avoid using multiple issuers with a particular origin.

tfpauly (approved):
Looking good, just some minor comments

PR #316, opened 2023-01-28
Use RATS terminology

### LINKED ISSUE #302: Improve attestation terminology
After talking with @hannestschofenig and @thomas-fossati, it seems like we could benefit from more clarity around attestation. In particular, we could improve clarity by providing a mapping between the terminology used in the architecture document and the terminology used in the RATS working group (which already does attestation stuff). Moreover, we could provide more examples of how attestation might actually work in practice. Currently, the diagrams just have this opaque "Attest" message that's sent in the issuance protocol, but expanding on that with details and an example or two could help align everyone's mental model.

chris-wood:
Closes #302

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -351,7 +351,7 @@ attestation is the process by which an Attester bears witness to, confirms,
 or authenticates a Client so as to verify a property about the Client that
 is required for Issuance. {{?RFC9334}} describes an architecture for attestation
 procedures. Using that architecture as a conceptual basis, Clients are
-attesters that produce attestation evidence, and Attesters are verififiers that
+RATS attesters that produce attestation evidence, and Attesters are RATS verififiers that
 appraise the validity of attestation evidence.
 
 The type of attestation procedure is a deployment-specific option and outside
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -346,23 +346,30 @@ See {{attester}} for more details.
...
-- Trusted device. Some Clients run on trusted hardware that are capable of
-  producing device-level attestation statements.
-
-Each of these attestation types has different security properties. For example,
+Attestation is an important part of the issuance protocol. In Privacy Pass,
+attestation is the process by which an Attester bears witness to, confirms,
+or authenticates a Client so as to verify a property about the Client that
+is required for Issuance. {{?RFC9334}} describes an architecture for attestation
+procedures. Using that architecture as a conceptual basis, Clients are
+attesters that produce attestation evidence, and Attesters are verififiers that
```
tfpauly, line 354: I wonder if our Attester also shares some of the RATS Endorser role in this case?

I also think we should disambiguate with adding RATS in the names
```suggestion
RATS attesters that produce attestation evidence, and Attesters are RATS verififiers that
```

chris-wood: I think that for the current PAT deployment the attester is also an endorser, but it's not clear to me if that's generally true for all types of attestation procedures.

PR #336, opened 2023-02-23
Not specific to HTTP

### LINKED ISSUE #323: Does the architecture need to use HTTP?
The interaction model here doesn't really need to use HTTP as far as I can tell.  Obviously, the specific protocols need to be more concrete, but the architecture works without specifically saying that the interactions use HTTP.

The overview specifically mentions HTTP only with respect to the Client-Origin interactions, not other interactions, but I can't see how this is necessary.

chris-wood:
Closes #323


tfpauly (approved):
Request/response is still a bit implying HTTP, but I'm OK with this change.

PR #339, opened 2023-02-23
Discuss what happens to tokens in circulation after compromise

### LINKED ISSUE #329: Broken attestations are forever
Section 3.2.1 talks about the need for issuers to trust attesters.  This is probably the main trust relationship (see #328) involved regarding attestation.  The other trust basis is the extent to which the attester trusts the client, though the point is probably that the attester starts out being extremely skeptical about the client and so acts to validate them.

This is fine (though I would delete the first sentence of the last paragraph of this section).  However, the remainder of this text talks about how to handle an attester that goes bad.  None of this acknowledges the real problem, which is that in a setting where an issuer works with multiple attesters, but issues from a single key, a failure in one attester cannot result in tokens being invalidated for just that attester.  The information about which attester was involved in token issuance is erased by the issuer, deliberately.

It might even be the case (thanks to consistency requirements) that the issuer cannot revoke tokens at all.  The only recourse is the one that is noted, which is that the attester cannot be used for future issuance.  The doc should probably talk about those tokens that were issued though.

chris-wood:
Closes #329

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -439,18 +439,28 @@ trusted device is less than the number of Clients that have solved a CAPTCHA in
 the past day. Attesters SHOULD not admit attestation types that result in small
 anonymity sets.
 
-The trustworthiness of Attesters depends on their ability to correctly and
-reliably perform attestation during the issuance protocol. Indeed, Issuers
-trust Attesters to correctly and reliably perform attestation. However, certain
-types of attestation can vary in value over time, e.g., if the attestation
-process is compromised or maliciously automated. These are considered
-exceptional events and require configuration changes to address the underlying
-cause. For example, if attestation is compromised because of a zero-day exploit
-on compliant devices, then the corresponding attestation procedure should be
-untrusted until the exploit is patched. Addressing changes in attestation
-quality is therefore a deployment-specific task. In Split Attester and Issuer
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -439,18 +439,28 @@ trusted device is less than the number of Clients that have solved a CAPTCHA in
...
-exceptional events and require configuration changes to address the underlying
-cause. For example, if attestation is compromised because of a zero-day exploit
-on compliant devices, then the corresponding attestation procedure should be
-untrusted until the exploit is patched. Addressing changes in attestation
-quality is therefore a deployment-specific task. In Split Attester and Issuer
-deployments (see {{deploy-split}}), Issuers can choose to remove compromised
-Attesters from their trusted set until the compromise is patched.
+Issuers trust Attesters to correctly and reliably perform attestation. However,
+certain types of attestation can vary in value over time, e.g., if the
+attestation procedure is compromised or maliciously automated. Broken
```
tfpauly, line 444: What does "maliciously automated" mean? You could just remove that:
```suggestion
attestation procedure is compromised. Broken
```

tfpauly (approved):
Approved pending one comment

draft-ietf-privacypass-architecture.md
chris-wood: This was existing text, but removing it is better!

PR #342, opened 2023-02-23
Explicit trust relationships (and privacy considerations)

### LINKED ISSUE #326: Cardinality of different relationships
From a bullet point in Section 3.1:

> Each Issuer, in turn, determines which Attesters it is willing to accept in the issuance protocol. This means that if an Origin origin.example accepts tokens issued by Issuer issuer.example, and that Issuer in turn accepts different types of attestation from more than one trusted Attester, then a Client may use either of these trusted Attesters to issue and redeem tokens for origin.example. 

It seems like the challenge points to a single issuer and the issuer points to one or more attesters.  If this is the case, then it might be worth spending a paragraph or sentence on it rather than burying it in an example.  Also, at this stage, it is not clear why the Origin couldn't point to more than one Issuer, if it were willing to do so.  A forward reference to discussions about this limit here being necessary to prevent widening the cross-contextual flow of information; though I believe that if the client were to be consistent in its choice, there is no additional information leak (other than perhaps something inherent to the client's choice).

### LINKED ISSUE #333: Authenticating the attester
In Section 4.3 (Join Origin and Issuer):

> Since the Attester and Issuer are separate entities, the Attester must authenticate itself to the Issuer. In settings where the Attester is a Client-trusted service, one way Attesters can authenticate to Issuers is via mutually-authenticated TLS. 

This appears to be a more general statement that could exist outside of where it appears.  There are lots of of things that relate to the (pairwise) trust relationships between Origin, Attester, and Issuer.

Furthermore, mutual authentication applies only to the extent that there is direct communication between entities.  There is a need for certain pieces of information that passes between these entities to be authenticated, but it would be better to articulate what that data is than to talk about the mechanism so directly.  That is, start by saying that when the issuer issues a token, it needs to be assured by the attester (one that it trusts) that the nonce it is binding the token was provided by a client that met whatever requirements the attester and issuer agreed would be necessary for issuance.

One way to ensure that the issuer can be sure that the attester did this is to have the attester present a request to issue a token itself, over a connection where the attester authenticates itself.  An alternative, that might be worse in that it uses bearer tokens, is to rely on the attester signing a claim to that effect and having some other entity (like the client) request issuance.

chris-wood:
Closes #326 
Closes #333


draft-ietf-privacypass-architecture.md
```diff
@@ -147,9 +147,13 @@ Attestation procedure:
...
 The Privacy Pass architecture consists of four logical entities --
 Client, Origin, Issuer, and Attester -- that work in concert
-for token redemption and issuance. This section describes the purpose
-of token the redemption and issuance protocols and the requirements
-on the relevant participants.
+for token redemption and issuance. This section presents an overview
+of Privacy Pass, a high-level description of the threat model and
+privacy goals of the architecture, and the purpose of the the redemption
+and issuance protocols, including requirements on the relevant participants
+in these protocols, necessary to achieve the intended privacy goals.
```
tfpauly, line 154:
```suggestion
privacy goals of the architecture, and the goals and requirements of the
redemption and issuance protocols.
```

draft-ietf-privacypass-architecture.md
```diff
@@ -229,12 +235,30 @@ Attestation context:
...
 including information needed for the attestation procedure to complete.
 
-The privacy goals of Privacy Pass are oriented around unlinkability based on
-these contexts. In particular, Privacy Pass aims to achieve three different
-types of unlinkability:
+The privacy goals of Privacy Pass assume a threat model in which Origins trust
+specific Issuers to produce tokens, and Issuers in turn trust one or more
+Attesters to correctly run the attestation procedure with Clients. This
+arrangement ensures that tokens which validate for a given Issuer were only
+issued to a Client that successfully complete attestation with an Attester that
```
tfpauly, line 245:
```suggestion
issued to a Client that successfully completed attestation with an Attester that
```

draft-ietf-privacypass-architecture.md
```diff
@@ -229,12 +235,30 @@ Attestation context:
...
+the Issuer trusts. Moreover, this arrangement means that if an Origin accepts
+tokens issued by an Issuer that trusts multiple Attesters then a Client can
```
tfpauly, line 247:
```suggestion
tokens issued by an Issuer that trusts multiple Attesters, then a Client can
```

draft-ietf-privacypass-architecture.md
```diff
@@ -229,12 +235,30 @@ Attestation context:
...
+use any of these Attesters to issue and redeem tokens for the Origin.
```
tfpauly, line 248:
```suggestion
use any one of these Attesters to issue and redeem tokens for the Origin.
```

draft-ietf-privacypass-architecture.md
```diff
@@ -229,12 +235,30 @@ Attestation context:
...
+
+The mechanisms for establishing trust between each entity in this arrangement
+are deployment specific. For example, in settings where Attesters and Issuers
```
tfpauly, line 251:
```suggestion
are deployment-specific. For example, in settings where Attesters and Issuers
```

draft-ietf-privacypass-architecture.md
```diff
@@ -336,6 +356,12 @@ redeemed already in those redemption contexts, since these tokens can
 be issued and then spent multiple times in response to any such challenge.
 See {{Section 2.1.1 of AUTHSCHEME}} for discussion.
 
+How Clients respond to token challenges can have privacy implications.
+For example, if a token challenge admits a choice of Issuer, then the choice
```
tfpauly, line 360: What does it mean for a "token challenge to admit a choice of Issuer"?

draft-ietf-privacypass-architecture.md
```diff
@@ -623,11 +649,12 @@ name.
...
-separate entities, the Attester must authenticate itself to the Issuer. In
-settings where the Attester is a Client-trusted service, one way Attesters
-can authenticate to Issuers is via mutually-authenticated TLS. However,
-alernative authentication mechanisms are possible. This arrangement is shown
-below.
+separate entities, this model requires some mechanism by which Issuers
+establish trust in the Attester (as described in {{privacy-and-trust}}).
+For example, in settings where the Attester is a Client-trusted service that
+directly communicates with the Issuer, one way to establish this trust is via
+mutually-authenticated TLS. However, alernative authentication mechanisms are
```
tfpauly, line 656:
```suggestion
mutually-authenticated TLS. However, alternative authentication mechanisms are
```

draft-ietf-privacypass-architecture.md
chris-wood: As a hypothetical, imagine a token challenge specified more than one issuer, or if clients were given two challenges for two different issuers.

tfpauly: I see. Our current challenge does not allow this to be expressed. How about:
```suggestion
For example, if an Origin allows the Client to choose an Issuer, then the choice
```

tfpauly (approved):

PR #343, opened 2023-02-23
More clear context descriptions

### LINKED ISSUE #325: Contexts and unlinkability goals
The definition of the redemption/issuance/attestation contexts seem to exist primarily to support the different unlinkability goals in the document.  Let's take the redemption one, which is defined as:

> The interactions and set of information shared between the Client and Origin. This context includes all information associated with redemption, such as the timestamp of the event, Client visible information (including the IP address), and the Origin name.

<details>
<summary>A digression on perspective</summary>

Note that "Client visible" is probably ambiguous here; it could mean information about the client that is visible (though to whom?) or information that is visible to the client.

This suggests that your goal is establish a point of reference or perspective for these: in this case, the relevant perspective is that of an Origin.  But this was not my main point...
</details>

The corresponding unlinkability definition says:

> Origin-Client unlinkability. This means that given two redemption contexts, the Origin cannot determine if both redemption contexts correspond to the same Client or two different Clients. Informally, this means that a Client in a redemption context is indistinguishable from any other Client that might use the same redemption context. The set of Clients that share the same redemption context is referred to as a redemption anonymity set.

This is a little confusing, because you might conceive of the defines the context as enclosing a single interaction between the Client and Origin, but this implies that a context encompasses multiple interactions with either the same client or different clients (all of which are indistinguishable).  That makes the definition of the context seem like it lacks something.  Does this refer to the information that is provided within that context that might be used to identify a client, from the perspective of the server?

Finally, there is mention of an "an empty redemption context" in Section 3.1.  I had no idea how to interpret this claim.  Does empty refer to the set of clients? The set of interactions?  The information provided in that context?

chris-wood:
Closes #325

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -222,7 +222,9 @@ and Issuer, i.e., the information that is provided or otherwise available
 to Attester and Issuer during issuance that might be used to identify a Client.
 This context includes all information associated with issuance, such as the
 timestamp of the event, any Client visible information (including the IP
-address), and the Origin name (if revealed during issuance).
+address), and the Origin name (if revealed during issuance). This does not include
+the token challenge in its entirety, as that is kept secret from the Issuer during the
+issuance protocol.
 
 Attestation context:
 : The interactions and set of information shared between
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -212,7 +212,7 @@ Redemption context:
 : The interactions and set of information shared
 between the Client and Origin, i.e., the information that is provided or
 otherwise available to the Origin during redemption that might be used
-to identify a Client. This context includes all information associated
+to identify a Client and construct a token challenge. This context includes all information associated
 with redemption, such as the timestamp of the event, Client visible
 information (including the IP address), and the Origin name.
 
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -210,23 +210,28 @@ contexts:
 
 Redemption context:
 : The interactions and set of information shared
-between the Client and Origin. This context includes all information
-associated with redemption, such as the timestamp of the event, Client
-visible information (including the IP address), and the Origin name.
+between the Client and Origin, i.e., the information that is provided or
+otherwise available to the Origin during redemption that might be used
+to identify a Client. This context includes all information associated
+with redemption, such as the timestamp of the event, Client visible
+information (including the IP address), and the Origin name.
 
 Issuance context:
-: The interactions and set of information shared
-between the Client, Attester, and Issuer. This context includes all
-information associated with issuance, such as the timestamp of the event,
-any Client visible information (including the IP address), and the
-Origin name (if revealed during issuance).
+: The interactions and set of information shared between the Client, Attester,
+and Issuer, i.e., the information that is provided or otherwise available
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -210,23 +210,28 @@ contexts:
 
 Redemption context:
 : The interactions and set of information shared
-between the Client and Origin. This context includes all information
-associated with redemption, such as the timestamp of the event, Client
-visible information (including the IP address), and the Origin name.
+between the Client and Origin, i.e., the information that is provided or
+otherwise available to the Origin during redemption that might be used
+to identify a Client. This context includes all information associated
+with redemption, such as the timestamp of the event, Client visible
```
martinthomson, line 216: and the challenge too?

draft-ietf-privacypass-architecture.md
```diff
@@ -210,23 +210,28 @@ contexts:
...
-between the Client, Attester, and Issuer. This context includes all
-information associated with issuance, such as the timestamp of the event,
-any Client visible information (including the IP address), and the
-Origin name (if revealed during issuance).
+: The interactions and set of information shared between the Client, Attester,
+and Issuer, i.e., the information that is provided or otherwise available
+to Attester and Issuer during issuance that might be used to identify a Client.
+This context includes all information associated with issuance, such as the
+timestamp of the event, any Client visible information (including the IP
+address), and the Origin name (if revealed during issuance).
```
martinthomson, line 225: This does not include the challenge, right?  It includes configuration information from the challenge (the type of issuance protocol), but not the challenge content itself, which needs to remain secret.

martinthomson (approved):

draft-ietf-privacypass-architecture.md
chris-wood: The Origin constructs that, so I think it's sort of implicit, but we can add that too.

draft-ietf-privacypass-architecture.md
```diff
@@ -210,23 +210,28 @@ contexts:
 
 Redemption context:
 : The interactions and set of information shared
-between the Client and Origin. This context includes all information
-associated with redemption, such as the timestamp of the event, Client
-visible information (including the IP address), and the Origin name.
+between the Client and Origin, i.e., the information that is provided or
+otherwise available to the Origin during redemption that might be used
+to identify a Client. This context includes all information associated
```
chris-wood, line 215:
```suggestion
to identify a Client and construct a token challenge. This context includes all information associated
```

draft-ietf-privacypass-architecture.md
```diff
@@ -210,23 +210,28 @@ contexts:
...
-between the Client, Attester, and Issuer. This context includes all
-information associated with issuance, such as the timestamp of the event,
-any Client visible information (including the IP address), and the
-Origin name (if revealed during issuance).
+: The interactions and set of information shared between the Client, Attester,
+and Issuer, i.e., the information that is provided or otherwise available
+to Attester and Issuer during issuance that might be used to identify a Client.
+This context includes all information associated with issuance, such as the
+timestamp of the event, any Client visible information (including the IP
+address), and the Origin name (if revealed during issuance).
```
chris-wood, line 225:
```suggestion
address), and the Origin name (if revealed during issuance). This does not include
the token challenge in its entirety, as that is kept secret from the Issuer during the
issuance protocol.
```

tfpauly (approved):

PR #346, opened 2023-02-27
Information flow

### LINKED ISSUE #332: Can the Origin distinguish per- and cross-origin tokens?
It seems like this is not the case (from Section 4.3):

> If there is only a single Origin, then per-Origin tokens are not appropriate in this model, since the Attester can learn the redemption context. However, the Attester does not learn whether a token is per-Origin or cross-Origin.

This implies something about the way in which per-origin tokens are attested that does not follow from any of the previous text.  But I don't know what.  Is it the case that the client needs to communicate the origin that a per-origin token needs to be bound to?  Does this assume that the issuer in this deployment configuration is receiving this information via the attester?  Why is it that the client can't use a proxy (or OHTTP) to request a token from the issuer?

chris-wood:
Closes #332

cc @martinthomson, this is largely inspired by your feedback! The intent here is to explain roughly how information flows throughout the various parties in the system, and then to lean on that flow of information in the deployment model section.

### DIFF
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -551,7 +551,7 @@ specified in {{issuer-role}} for managing Issuer public key data.
 
 ## Information Flow {#flow}
 
-The end-to-end run of redemption and issuance protocol involves information
+The end-to-end process of redemption and issuance protocols involves information
 flowing between Issuer, Origin, Client, and Attester. That information can
 have implications on the privacy goals that Privacy Pass aims to provide
 as outlined in {{privacy-and-trust}}. In this section, we describe the flow
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -175,7 +175,8 @@ CAPTCHA, checking device or hardware attestation validity, etc; see
 {{attester}} for more details.
 
 4. If the attestation process completes successfully, the client creates a
-Token Request to send to the designated Issuer (generally via the Attester).
+Token Request to send to the designated Issuer (generally via the Attester,
+though it is not required to be sent through the Attester).
 The Attester and Issuer might be functions on the same server, depending on the
 deployment model (see {{deployment}}). Depending on the attestation process, it
 is possible for attestation to run alongside the issuance protocol, e.g., where
@@ -248,9 +249,11 @@ tokens issued by an Issuer that trusts multiple Attesters then a Client can
 use any of these Attesters to issue and redeem tokens for the Origin.
 
 The mechanisms for establishing trust between each entity in this arrangement
-are deployment specific. For example, in settings where Attesters and Issuers
-communicate over TLS, Attesters and Issuers may use mutually authenticated TLS
-to authenticate one another.
+are deployment specific. For example, in settings where Clients interact with
+Issuers transitively through an Attester, Attesters and Issuers might use
+mutually authenticated TLS to authenticate one another. In settings where
+Clients do not communicate with Issuers through an Attester, the Attesters
+might convey this trust via a digital signature over that Issuers can verify.
 
 Clients explicitly trust Attesters to perform attestation correctly and in a
 way that does not violate their privacy. However, Clients assume Issuers and
@@ -603,7 +606,8 @@ or issuance contexts as described below.
 - Origin information. The issuance protocol can contribute information about
 the Origin in how it responds to a token request. For example, if an Issuer
 learns the Origin during issuance and is also configured to respond in some way
-on the basis of that information, that response can reveal information to the
+on the basis of that information, and the Client interacts with the Issuer
+transitively through the Attester, that response can reveal information to the
 Attester.
 - Token. The token produced by the issuance protocol can contain information
 from the issuance context. In particular, depending on the issuance protocol,
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -561,7 +561,7 @@ goals in particular deployment models is further discussed in {{deployment}}.
 
 To use Privacy Pass, Origins choose an Issuer from which they are willing to
 accept tokens. Origins then construct a token challenge using this specified
-Issuer and the redemption context it shares with the Client. This token
+Issuer and information from the redemption context it shares with the Client. This token
 challenge is then delivered to a Client. The token challenge conveys
 information about the Issuer and the redemption context, such as whether the
 Origin desires a per-Origin or cross-Origin token. Any entity that sees
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -576,9 +576,10 @@ or issuance contexts as described below.
 
 - Issuance protocol. The type of issuance protocol can contribute information
 about the Issuer's capabilities to the attestation or issuance contexts.
-- Issuer identity or configuration. Information about the Issuer, such as its
-identity or the public key used to validate tokens it creates, can be revealed
-during issuance and contribute to the attestation or issuance contexts.
+- Issuer configuration. Information about the Issuer configuration, such as
+its identity or the public key used to validate tokens it creates, can be
+revealed during issuance and contribute to the attestation or issuance
+contexts.
 - Attestation information. The issuance protocol can contribute information to
 the issuance context based on what attestation procedure the Issuer uses to
 trust a token request. In particular, a token request that is validated by
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -251,7 +251,7 @@ use any of these Attesters to issue and redeem tokens for the Origin.
 
 The mechanisms for establishing trust between each entity in this arrangement
 are deployment specific. For example, in settings where Clients interact with
-Issuers transitively through an Attester, Attesters and Issuers might use
+Issuers through an Attester, Attesters and Issuers might use
 mutually authenticated TLS to authenticate one another. In settings where
 Clients do not communicate with Issuers through an Attester, the Attesters
 might convey this trust via a digital signature over that Issuers can verify.
--- a/draft-ietf-privacypass-architecture.md
+++ b/draft-ietf-privacypass-architecture.md
@@ -235,9 +235,10 @@ Origin name (if revealed during issuance).
 Attestation context:
 : The interactions and set of information shared between
 the Client and Attester only, for the purposes of attesting the vailidity of
-the Client. This context includes all information associated with attestation,
-such as the timestamp of the event and any Client visible information,
-including information needed for the attestation procedure to complete.
+the Client for the purposes of completing the issuance protocol. This context
+includes all information associated with attestation, such as the timestamp of
+the event and any Client visible information, including information needed for
+the attestation procedure to complete.
 
 The privacy goals of Privacy Pass assume a threat model in which Origins trust
 specific Issuers to produce tokens, and Issuers in turn trust one or more
@@ -569,13 +570,22 @@ the token challenge might learn things about the Client as known to the Origin.
 This is why input secrecy is a requirement for issuance protocols, as it
 ensures that the challenge is not directly available to the Issuer.
 
-### Attestation and Issuance Flow {#issue-flow}
+### Attestation Flow {#attestation-flow}
+
+Clients interact with the Attester to prove that they meet some required
+set of properties. In doing so, Clients contribute information to the
+attestation context, which might include sensitive information such as
+application-layer identities, IP addresses, and so on. Clients can choose
+whether or not to contribute this information based on local policy or
+preference.
+
+### Issuance Flow {#issue-flow}
 
 Clients use the issuance protocol to produce a token bound to a token
 challenge. In doing so, there are several ways in which the issuance protocol
-contributes information to the attestation context and issuance context.
-For example, a token request may contribute information to the attestation
-or issuance contexts as described below.
+contributes information to the attestation or issuance contexts. For example, a
+token request may contribute information to the attestation or issuance
+contexts as described below.
 
 - Issuance protocol. The type of issuance protocol can contribute information
 about the Issuer's capabilities to the attestation or issuance contexts, as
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+- Issuance protocol. The type of issuance protocol can contribute information
+about the Issuer's capabilities to the attestation or issuance contexts.
+- Issuer identity or configuration. Information about the Issuer, such as its
+identity or the public key used to validate tokens it creates, can be revealed
+during issuance and contribute to the attestation or issuance contexts.
+- Attestation information. The issuance protocol can contribute information to
+the issuance context based on what attestation procedure the Issuer uses to
+trust a token request. In particular, a token request that is validated by
+a given Attester means that the Client which generated the token request must
+be capable of the completing the designated attestation procedure.
```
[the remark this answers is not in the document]
martinthomson: This is necessary information, isn't it?  We really do need the Issuer to be convinced that the Client passed whatever tests the Attester set.

The other aspect to consider is what unwanted information might be revealed here.  If the Attester would accept a Client that meets criterion A, B, or C, then it should not be the case that the Issuer learns which of A, B, or C was used.  That seems fine.  However, if there are multiple valid Attesters, I don't see a way to avoid the Issuer learning which Attester was used.  The choice of Attester therefore says something about the client.  There is a transitory flow of information here though: if Attester A supports method A and Attester B supports method B, then the choice of Attester indirectly reveals the choice of method.

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+- Origin information. The issuance protocol can contribute information about
+the Origin that challenged the Client in {{challenge-flow}}. In particular,
+a token request designated for a specific Issuer means that the resulting token
+is for an Origin that trusts the specified Issuer.
```
martinthomson, line 587: I think that you might infer that the Client had received a challenge from one of the Origins that trust that Issuer, but you can't guarantee that this is the case.  Clients could proactively retrieve tokens.  Those tokens wouldn't be bound to something the Origin cares about necessarily, so maybe a per-Origin token doesn't fit this cleanly, but it does seem possible.

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+
+Moreover, a token response may contribute information to the attestation
+or issuance contexts as described below.
+
+- Origin information. The issuance protocol can contribute information about
+the Origin in how it responds to a token request. For example, if an Issuer
+learns the Origin during issuance and is also configured to respond in some way
+on the basis of that information, that response can reveal information to the
+Attester.
```
martinthomson, line 596: I'll repeat my previous complaint.  The Client interacting with the Issuer *via the Attester* is something that is assumed far too often in the document.  This needs to be conditional at a minimum.  Also, in those protocols that encapsulate the Client-Issuer exchange, the information that leaks is only based on traffic analysis.

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+### Token Challenge Flow {#challenge-flow}
+
+To use Privacy Pass, Origins choose an Issuer from which they are willing to
+accept tokens. Origins then construct a token challenge using this specified
+Issuer and the redemption context it shares with the Client. This token
+challenge is then delivered to a Client. The token challenge conveys
+information about the Issuer and the redemption context, such as whether the
+Origin desires a per-Origin or cross-Origin token.
+
+### Attestation and Issuance Flow {#issue-flow}
```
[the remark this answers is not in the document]
chris-wood: Yeah, I think so, since they're both happening in issuance.

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+have implications on the privacy goals that Privacy Pass aims to provide
+as outlined in {{privacy-and-trust}}. In this section, we describe the flow
+of information between each party. How this information affects the privacy
+goals in particular deployment models is further discussed in {{deployment}}.
+
+### Token Challenge Flow {#challenge-flow}
+
+To use Privacy Pass, Origins choose an Issuer from which they are willing to
+accept tokens. Origins then construct a token challenge using this specified
+Issuer and the redemption context it shares with the Client. This token
```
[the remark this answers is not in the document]
chris-wood: I think we need to keep redemption context here, since that's the information the Origin has available to it to construct a token challenge. I like the other suggestion to explain why issuance secrecy is important.

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+challenge. In doing so, there are several ways in which the issuance protocol
+contributes information to the attestation context and issuance context.
+For example, a token request may contribute information to the attestation
+or issuance contexts as described below.
+
+- Issuance protocol. The type of issuance protocol can contribute information
+about the Issuer's capabilities to the attestation or issuance contexts.
+- Issuer identity or configuration. Information about the Issuer, such as its
+identity or the public key used to validate tokens it creates, can be revealed
+during issuance and contribute to the attestation or issuance contexts.
```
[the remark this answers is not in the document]
chris-wood: I relaxed this to be more general and just say "issuer configuration," as that's generally what the token challenge conveys.

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
chris-wood: This list isn't meant to be exhaustive, as I don't expect us to be able to list all exhaustive things. It contains the what seem to me to be the obvious things that might be revealed, and the type of attestation that's done is one of those obvious things. It is necessarily revealed to the issuer through the process here.

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +545,76 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+### Attestation and Issuance Flow {#issue-flow}
+
+Clients use the issuance protocol to produce a token bound to a token
+challenge. In doing so, there are several ways in which the issuance protocol
+contributes information to the attestation context and issuance context.
+For example, a token request may contribute information to the attestation
+or issuance contexts as described below.
+
+- Issuance protocol. The type of issuance protocol can contribute information
+about the Issuer's capabilities to the attestation or issuance contexts.
```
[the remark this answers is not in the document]
chris-wood: > Why would the issuance protocol choice leak information to the issuance context? That's not immediately obvious.

As an example, maybe an Issuer only supports the rate-limiting issuance protocol for one origin. This would be pretty silly, but it would be possible, and as a result, the issuance protocol immediately reveals unwanted information to the issuer.

> I think that what you are saying is that the choice made by the Origin regarding the issuance protocol might be used to carry information from the Origin to the Issuer if the Client requests issuance when it receives a challenge. If the issuance protocol choice is exposed to the Attester, then it too learns this information. If there are multiple options available, some information about the Client can be encoded in the choice between those options.

That's a good clarification, yeah. I'll include it.

draft-ietf-privacypass-architecture.md
chris-wood: Good call out. I've clarified that it's only a possibility.

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
martinthomson: I still don't understand whether "redemption context" refers to data or something else though.

draft-ietf-privacypass-architecture.md
chris-wood: This is fair criticism. It's noted that this is an option in the overview section, but I don't think it's obvious enough. I'll try to clarify.

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
chris-wood: It refers to _anything that the origin can see during redemption_ (defined earlier in the doc).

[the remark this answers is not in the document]
chris-wood: (Unresolving) @martinthomson does the improved description in #343 work?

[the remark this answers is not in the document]
martinthomson:
```suggestion
Issuer and information from the redemption context it shares with the Client. This token
```

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +548,88 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+Origin desires a per-Origin or cross-Origin token. Any entity that sees
+the token challenge might learn things about the Client as known to the Origin.
+This is why input secrecy is a requirement for issuance protocols, as it
+ensures that the challenge is not directly available to the Issuer.
+
+### Attestation and Issuance Flow {#issue-flow}
+
+Clients use the issuance protocol to produce a token bound to a token
+challenge. In doing so, there are several ways in which the issuance protocol
+contributes information to the attestation context and issuance context.
```
martinthomson, line 576: I'm still not convinced by this being a single section.  The attestation context only gains information from the issuance context transitively.  That is, if you need to use attestation method W to execute issuance protocol X and you need issuance protocol X to get token Y to answer challenge Z, then you can trace the way from Z to W, but each step is its own and the client is more or less in control of whether to proceed in each, provided that it understands what information carries over.

draft-ietf-privacypass-architecture.md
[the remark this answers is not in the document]
martinthomson: Obvious or not, it is still worth spelling out explicitly.  These are points where information leaks as a direct result of the architectural choices.

draft-ietf-privacypass-architecture.md
chris-wood: The attestation context does sometimes gain information explicitly from issuance, as in the rate limited protocol. I'll try to clarify that in the definitions.

draft-ietf-privacypass-architecture.md
```diff
@@ -234,9 +235,10 @@ Origin name (if revealed during issuance).
 Attestation context:
 : The interactions and set of information shared between
 the Client and Attester only, for the purposes of attesting the vailidity of
-the Client. This context includes all information associated with attestation,
-such as the timestamp of the event and any Client visible information,
-including information needed for the attestation procedure to complete.
+the Client for the purposes of completing the issuance protocol. This context
```
tfpauly, line 238: Two "for the purposes of..." here?

draft-ietf-privacypass-architecture.md
```diff
@@ -248,9 +250,11 @@ tokens issued by an Issuer that trusts multiple Attesters then a Client can
 use any of these Attesters to issue and redeem tokens for the Origin.
 
 The mechanisms for establishing trust between each entity in this arrangement
-are deployment specific. For example, in settings where Attesters and Issuers
-communicate over TLS, Attesters and Issuers may use mutually authenticated TLS
-to authenticate one another.
+are deployment specific. For example, in settings where Clients interact with
+Issuers transitively through an Attester, Attesters and Issuers might use
```
tfpauly, line 254:
```suggestion
Issuers through an Attester, Attesters and Issuers might use
```

draft-ietf-privacypass-architecture.md
```diff
@@ -544,6 +549,101 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
 Any extension to the Privacy Pass protocol MUST adhere to the guidelines
 specified in {{issuer-role}} for managing Issuer public key data.
 
+## Information Flow {#flow}
+
+The end-to-end run of redemption and issuance protocol involves information
```
tfpauly, line 554:
```suggestion
The end-to-end process of redemption and issuance protocols involves information
```

tfpauly (approved):

chris-wood:
@martinthomson I think we've faithfully responded to your helpful comments and review, so I'm going to merge this. Please do let us know if you think there's more that should be done, though!

draft-ietf-privacypass-architecture.md
```diff
@@ -566,6 +570,101 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+Clients use the issuance protocol to produce a token bound to a token
+challenge. In doing so, there are several ways in which the issuance protocol
+contributes information to the attestation or issuance contexts. For example, a
+token request may contribute information to the attestation or issuance
+contexts as described below.
+
+- Issuance protocol. The type of issuance protocol can contribute information
+about the Issuer's capabilities to the attestation or issuance contexts, as
+well as the capabilities of a given Client. For example, if a Client is
+presented with multuiple issuance protocol options, then the choice of which
```
martinthomson, line 614: typo: mult~~u~~iple

draft-ietf-privacypass-architecture.md
```diff
@@ -566,6 +570,101 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+issuance protocol to use can contrbute information about the Client's
```
martinthomson, line 615: typo: contr[i]bute

draft-ietf-privacypass-architecture.md
```diff
@@ -566,6 +570,101 @@ and guidelines on how to deploy the protocol to minimize any privacy impacts.
...
+learns the Origin during issuance and is also configured to respond in some way
+on the basis of that information, and the Client interacts with the Issuer
+transitively through the Attester, that response can reveal information to the
+Attester.
+- Token. The token produced by the issuance protocol can contain information
+from the issuance context. In particular, depending on the issuance protocol,
+tokens can contain public or private metadata, and Issuers can choose that
+metadata on the basis of information in the issuance context.
+
+In general, exceptional cases in the issuance protocol, such as when either the
```
martinthomson, line 648: I suggest that you avoid saying "In general" before "exceptional".

martinthomson (commented):
Thanks for all the work on this.

PR #355, opened 2023-04-10
TokenChallenge validation steps

### LINKED ISSUE #353: Address Developer Ergonomics by providing UA feedback for malformed challenges
A common bug I have seen with engineering teams implementing PAT is that they don't pad base64url encoded challenges (see https://github.com/ietf-wg-privacypass/base-drafts/issues/352). This easily goes un-noticed and impacts adoptions. The root challenge is that automated integration tests with PAT aren't possible compounded by the lack of feedback interfaces from the UA to report of a malformed challenges. 

I created a [few tools](https://private-access-token.colinbendell.dev/) to [help test](https://private-access-token.colinbendell.dev/debug.html) for some of the common errors. However, malformed rejection from the UA is still challenging. 

For example, the current version of the spec requires base64url with padding, and with padding the fields should be quoted. A developer following this spec would run into errors today because the current implementation in iOS and macOS reject quoted fields. This causes many hours of developer time trying to triage and investigate these kinds of errors. 

It would help if there was a standard mechanism for a UA to clearly report if a challenge is being rejected because it was malformed. A few possible suggestions:
1. send an http header, or a variant of the Authorization header signalling this malformed challenge on the next request. For example `Authorization: PrivateToken error="Malformed challenge"`
2. integrate the spec with the [ReportingAPI](https://w3c.github.io/reporting/) allowing for out of band, and privacy preserving feedback to the origin.

Having such a feedback mechanism will also allow for easier UA adoption by enabling WPT tests.

chris-wood:
Closes #353.

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -225,13 +225,15 @@ WWW-Authenticate:
   PrivateToken challenge="abc...", token-key="123..."
 ~~~
 
-Upon receipt of this challenge, a client uses the message and keys in the
-issuance protocol indicated by the token_type. If the TokenChallenge has a
-token_type the client does not recognize or support, it MUST NOT parse or
-respond to the challenge. If the TokenChallenge contains a non-empty
-origin_info field, the client MUST validate that the name of the origin
-that issued the authentication challenge is included in the list of origin
-names; if validation fails, the client MUST NOT process or respond to the
+Upon receipt of this challenge, a client validates the TokenChallenge before
+responding to it. Validation requirements are as follows:
+
+- The TokenChallenge structure is well-formed;
+- The token_type is recognized and supported by by the client; and
+- If the origin_info field is non-empty, the name of the origin that issued the
+  authentication challenge is included in the list of origin names. 
+
+If validation fails, the client MUST NOT process or respond to the
 challenge. Clients MAY have further restrictions and requirements around
 validating when a challenge is considered acceptable or valid. For example,
 clients can choose to ignore challenges that list origin names for which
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -225,13 +225,15 @@ WWW-Authenticate:
...
-origin_info field, the client MUST validate that the name of the origin
-that issued the authentication challenge is included in the list of origin
-names; if validation fails, the client MUST NOT process or respond to the
+Upon receipt of this challenge, a client validates the TokenChallenge before
+responding to it. Validation requirements are as follows:
+
+- The TokenChallenge structure is well-formed;
+- The token_type is recognized and supported by by the client; and
+- If the origin_info field is non-empty, the name of the origin that issued the
+  authentication challenge is included in the list of origin names.
```
chris-wood, line 234:
```suggestion
  authentication challenge is included in the list of origin names.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -225,13 +225,15 @@ WWW-Authenticate:
...
-token_type the client does not recognize or support, it MUST NOT parse or
-respond to the challenge. If the TokenChallenge contains a non-empty
-origin_info field, the client MUST validate that the name of the origin
-that issued the authentication challenge is included in the list of origin
-names; if validation fails, the client MUST NOT process or respond to the
+Upon receipt of this challenge, a client validates the TokenChallenge before
+responding to it. Validation requirements are as follows:
+
+- The TokenChallenge structure is well-formed;
+- The token_type is recognized and supported by by the client; and
```
chris-wood, line 232:
```suggestion
- The token_type is recognized and supported by the client; and
```

smhendrickson (approved):

PR #360, opened 2023-05-08
Add HTTP header encoding test vectors

### LINKED ISSUE #352: Add base64 examples to test vectors
A common bug I have seen with engineering teams implementing PAT is that they don't pad base64url encoded challenges. This is compounded by the lack of feedback from the UA. However, in this case, the padding is not needed for successful base64 decode because the lengths are known. Additionally, introducing `=` padding requires additional quoting of the fields which adds both visual and non-value noise to the exchange. 

`WWW-Authenticate: PrivateToken challenge="abc123==", token-key="def456=="`
vs
`WWW-Authenticate: PrivateToken challenge=abc123, token-key=def456`

I can appreciate the desire to make the spec less ambiguous to reduce the burden on the UA. However, given that there are a limited number of UAs but millions of origins, there is a greater chance that origins will be unknowingly misconfigured. Therefore it would improve developer ergonomics for origins if the UA was expected to be more flexible in the decode.

chris-wood:
Closes #352

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -749,7 +749,7 @@ ee33fcfdd1104c665eb827c20621
 This section includes test vectors the contents of the HTTP authentication
 headers. Each test vector consists of one or more challenges that comprise
 a WWW-Authenticate header. For each challenge, the token-type, token-key,
-max-age, and token-challenge parametrs are listed. Each challenge also
+max-age, and token-challenge parameters are listed. Each challenge also
 includes an unknown (not specified) parameter that implementations are meant
 to ignore.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -731,3 +743,89 @@ d02f390f9bf5ef14730e0ee15aa92df7eeba1bd7c550a8184bee32ce66e6fb527
 17aa67da7e0ca32f4cdca9dec7130f861220ad4241ee0e33eb4a486a32f05af05
 ee33fcfdd1104c665eb827c20621
 ~~~
+
+## HTTP Header Test Vectors
+
+This section includes test vectors the contents of the HTTP authentication
+headers. Each test vector consists of one or more challenges that comprise
+a WWW-Authenticate header. For each challenge, the token-type, token-key,
+max-age, and token-challenge parametrs are listed. Each challenge also
```
tfpauly, line 752:
```suggestion
max-age, and token-challenge parameters are listed. Each challenge also
```

tfpauly (approved):

smhendrickson (approved):

PR #380, opened 2023-06-14
Apply edits based on bemasc's review

### LINKED ISSUE #372: Minor editorial comments on draft-ietf-privacypass-protocol
(As individual participant)

1. URI vs. URL
* The text uses a mix of "URI" and "URL".  It's not clear to me that the usage follows a clear pattern or logic.
2. "Issuer Request URL value (as an absolute or relative URL)"
* Relative to what?  (Presumably the location of the configuration object?)  Are you sure that allowing "../request" is [a good idea](https://owasp.org/www-community/attacks/Path_Traversal) here?  Perhaps you may want to restrict the allowed values in some way.
3. "Issuer directory resources ... are located at the well-known location /.well-known/token-issuer-directory"
* This sentence presumes that the Issuer is identified by an HTTP Origin.  I would appreciate a sentence explaining why this is so, as opposed to identifying the Issuer by a full URL (as is [generally recommended](https://www.rfc-editor.org/rfc/rfc9205.html#section-4.4.1-3)).  Perhaps it is related to the use of origin-scoped authentication tokens in the Issuer == Origin case?  Or is it just a convenience for compact identification of Issuers?  Is it mandatory, or can Issuers be identified by a URLs as well?  The phrasing "are located" suggests that this is always true, but the later text that the Issuer name is "typically a host name" suggests that other arrangements (which might not rely on .well-known) are also allowed.
4. "When use of an Issuer configuration results in token issuance failures, e.g., because the configuration information is too stale, the directory SHOULD be fetched and revalidated."
* How does one know that a failure is due to stale configuration?  Presumably the HTTP resource for the configuration is still fresh or we wouldn't be using it.

chris-wood:
Closes #372

@bemasc at this point I think we would need substantial interest in moving away from a .well-known, so the text here simply clarifies why it exists.

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -219,7 +219,7 @@ Consumers of the Issuer directory resource SHOULD follow the usual HTTP caching
 result in use of stale Issuer configuration information, whereas short
 lifetimes may result in decreased performance. When use of an Issuer
 configuration results in token issuance failures, e.g., because the
-Issuer directory resource is no longer valid and issuance requests using this
+Issuer has invalidated its directory resource before its expiration time and issuance requests using this
 configuration are unsuccessful, the directory SHOULD be fetched and revalidated.
 
 # Issuance Protocol for Privately Verifiable Tokens {#private-flow}
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -194,7 +194,7 @@ Issuer directory resources have the media type
 "application/token-issuer-directory" and are located at the well-known location
 /.well-known/token-issuer-directory; see {{wkuri-reg}} for the registration
 information for this well-known URI. The reason that this resource is located
-at a well-known URI is that Issuers are defined by their origin in TokenChallenge
+at a well-known URI is that Issuers are defined by an origin name in TokenChallenge
 structures; see {{Section 2.1 of AUTHSCHEME}}.
 
 The Issuer directory and Issuer resources SHOULD be available on the same domain. If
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -193,10 +193,12 @@ Altogether, the Issuer's directory could look like:
 Issuer directory resources have the media type
 "application/token-issuer-directory" and are located at the well-known location
 /.well-known/token-issuer-directory; see {{wkuri-reg}} for the registration
-information for this well-known URI.
+information for this well-known URI. The reason that this resource is located
+at a well-known URI is that Issuers are defined by their origin in TokenChallenge
```
tfpauly, line 197:
```suggestion
at a well-known URI is that Issuers are defined by an origin name in TokenChallenge
```

tfpauly (approved):

draft-ietf-privacypass-protocol.md
```diff
@@ -217,8 +219,8 @@ Consumers of the Issuer directory resource SHOULD follow the usual HTTP caching
 result in use of stale Issuer configuration information, whereas short
 lifetimes may result in decreased performance. When use of an Issuer
 configuration results in token issuance failures, e.g., because the
-configuration information is too stale, the directory SHOULD be fetched and
-revalidated.
+Issuer directory resource is no longer valid and issuance requests using this
```
bemasc, line 222: It sounds like you're saying that the directory resource is fresh but invalid?  If so, it would be clearer to make that explicit.
```suggestion
Issuer has invalidated its directory resource before its expiration time and issuance requests using this
```

bemasc (approved):

dvorak42 (approved):

draft-ietf-privacypass-protocol.md
chris-wood: Yes, exactly, thanks for the suggestion!

PR #381, opened 2023-06-15
Add "not-before" field

chris-wood:
Based on [mailing list discussion](https://mailarchive.ietf.org/arch/msg/privacy-pass/vrhCH9cR2Ru1wUd6WJ5MC45lPqM/).

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -203,9 +203,9 @@ Altogether, the Issuer's directory could look like:
  }
 ~~~
 
-Clients that use this directory resource before 1686850920881 in UNIX time would use the
+Clients that use this directory resource before 1686913811 in UNIX time would use the
 second key in the "token-keys" list, whereas Clients that use this directory after
-1686850920881 in UNIX time would use the first key in the "token-keys" list.
+1686913811 in UNIX time would use the first key in the "token-keys" list.
 
 Issuer directory resources have the media type
 "application/private-token-issuer-directory" and are located at the well-known location
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -179,7 +179,10 @@ Beyond staging keys with the "not-before" value, Issuers MAY advertise multiple
 Issuers indicate preference for which token key to use based on the order of
 keys in the list, with preference given to keys earlier in the list. Clients
 SHOULD use the first key in the "token-keys" list that either does not have a
-"not-before" value or has a "not-before" value in the past.
+"not-before" value or has a "not-before" value in the past. Origins can attempt
+to use any key in the "token-keys" list to verify tokens, starting with the most
+preferred key in the list. Trial verification like this can help deal with Client
+clock skew.
 
 Altogether, the Issuer's directory could look like:
 
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -193,7 +193,7 @@ Altogether, the Issuer's directory could look like:
       {
         "token-type": 2,
         "token-key": "MI...AB",
-        "not-before": 1686850920881,
+        "not-before": 1686913811,
       },
       {
         "token-type": 2,
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -168,7 +168,7 @@ defined in {{tokenkeys-values}}.
 {: #tokenkeys-values title="Issuer 'token-keys' object description'"}
 
 Each "token-keys" JSON object may also contain the optional field "not-before".
-The value of this field is the UNIX timestamp (number of milliseconds since
+The value of this field is the UNIX timestamp (number of seconds since
 January 1, 1970, UTC) at which the key can be used. If this field is present,
 Clients SHOULD NOT use a token key before this timestamp, as doing so may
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -167,10 +167,17 @@ defined in {{tokenkeys-values}}.
...
+The value of this field is the UNIX timestamp (number of milliseconds since
+January 1, 1970, UTC) at which the key can be used. If this field is present,
+Clients SHOULD NOT use a token key before this timestamp, as doing so may
+lead to issuance failures. The purpose of this field is to assist in scheduled
+key rotations.
+
+Beyond staging keys with the "not-before" value, Issuers MAY advertise multiple
+token-keys for the same token-type to facilitate key rotation. In this case,
+Issuers indicate preference for which token key to use based on the order of
+keys in the list, with preference given to keys earlier in the list.
```
dvorak42, line 180: For keys without not-before, wouldn't clients always select the first key? What's the value of having multiple non-not-before keys?

chris-wood: Yeah, if there are no not-before values, clients would pick the first in the list.

chris-wood: I clarified how clients would consume a list based on not-before and the sorting order.

draft-ietf-privacypass-protocol.md
```diff
@@ -167,10 +167,19 @@ defined in {{tokenkeys-values}}.
 | token-key    | The base64url encoding of the Public Key for use with the issuance protocol, including padding, represented as a JSON string ({{RFC8259, Section 7}}) |
 {: #tokenkeys-values title="Issuer 'token-keys' object description'"}
 
-Issuers MAY advertise multiple token-keys for the same token-type to
-support key rotation. In this case, Issuers indicate preference for which
-token key to use based on the order of keys in the list, with preference
-given to keys earlier in the list.
+Each "token-keys" JSON object may also contain the optional field "not-before".
```
tfpauly, line 170: Should the key also be defined in the table above?

chris-wood: I left it out of the table because that is a list of required fields, whereas this one is optional.

draft-ietf-privacypass-protocol.md
```diff
@@ -167,10 +167,19 @@ defined in {{tokenkeys-values}}.
...
+Clients SHOULD NOT use a token key before this timestamp, as doing so may
+lead to issuance failures. The purpose of this field is to assist in scheduled
+key rotations.
+
+Beyond staging keys with the "not-before" value, Issuers MAY advertise multiple
+"token-keys" for the same token-type to facilitate key rotation. In this case,
+Issuers indicate preference for which token key to use based on the order of
+keys in the list, with preference given to keys earlier in the list. Clients
+SHOULD use the first key in the "token-keys" list that either does not have a
+"not-before" value or has a "not-before" value in the past.
```
smhendrickson, line 182: Do you want to explicitly call out origin behavior, where they should attempt to verify using all present keys, or add some skew around the `not-before` for verification?

I'm worried an origin will use the not-before as a hard cutoff, and clients with positive skew will use this before origins are allowing verification on it.

chris-wood: @smhendrickson done! How does it look?

smhendrickson: LGTM

punkeel (approved):

draft-ietf-privacypass-protocol.md
```diff
@@ -167,10 +167,22 @@ defined in {{tokenkeys-values}}.
 | token-key    | The base64url encoding of the Public Key for use with the issuance protocol, including padding, represented as a JSON string ({{RFC8259, Section 7}}) |
 {: #tokenkeys-values title="Issuer 'token-keys' object description'"}
 
-Issuers MAY advertise multiple token-keys for the same token-type to
-support key rotation. In this case, Issuers indicate preference for which
-token key to use based on the order of keys in the list, with preference
-given to keys earlier in the list.
+Each "token-keys" JSON object may also contain the optional field "not-before".
+The value of this field is the UNIX timestamp (number of milliseconds since
```
chris-wood, line 171:
```suggestion
The value of this field is the UNIX timestamp (number of seconds since
```

draft-ietf-privacypass-protocol.md
```diff
@@ -181,6 +193,7 @@ Altogether, the Issuer's directory could look like:
       {
         "token-type": 2,
         "token-key": "MI...AB",
+        "not-before": 1686850920881,
```
chris-wood, line 196:
```suggestion
        "not-before": 1686913811,
```

draft-ietf-privacypass-protocol.md
```diff
@@ -190,6 +203,10 @@ Altogether, the Issuer's directory could look like:
  }
 ~~~
 
+Clients that use this directory resource before 1686850920881 in UNIX time would use the
+second key in the "token-keys" list, whereas Clients that use this directory after
+1686850920881 in UNIX time would use the first key in the "token-keys" list.
```
chris-wood, line 208:
```suggestion
Clients that use this directory resource before 1686913811 in UNIX time would use the
second key in the "token-keys" list, whereas Clients that use this directory after
1686913811 in UNIX time would use the first key in the "token-keys" list.
```

PR #387, opened 2023-06-21
Clients MAY ignore challenges for any reason (but let's keep some examples)

### LINKED ISSUE #385: Ignoring challenges
From @paulwouters

> Clients can ignore the challenge if

Why "can" and not MAY ?

chris-wood:
Closes #385

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -219,7 +219,7 @@ using an out-of-band mechanism.
 - "max-age", an optional parameter that consists of the number of seconds for
 which the challenge will be accepted by the origin.
 
-Clients MAY ignore the challenge if desired, e.g., because the token-key is
+Clients MAY ignore the challenge, e.g., because the token-key is
 invalid or otherwise untrusted.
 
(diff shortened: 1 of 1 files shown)


tfpauly (approved):
Not too sure about "if desired", but I couldn't come up with a better suggestion.

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -219,8 +219,8 @@ using an out-of-band mechanism.
 - "max-age", an optional parameter that consists of the number of seconds for
 which the challenge will be accepted by the origin.
 
-Clients can ignore the challenge if the token-key is invalid or otherwise
-untrusted.
+Clients MAY ignore the challenge if desired, e.g., because the token-key is
```
chris-wood, line 222:
```suggestion
Clients MAY ignore the challenge, e.g., because the token-key is
```

chris-wood:
> Not too sure about "if desired", but I couldn't come up with a better suggestion.

Yeah, I wasn't too sure about it either. I just dropped it. 👍

tfpauly:
Sold!

PR #390, opened 2023-06-21
Make clear that strings are hostnames

### LINKED ISSUE #383: issuer_name clarification
From @paulwouters

>  opaque issuer_name<1..2^16-1>

with:

> The string is prefixed with a 2-octet integer indicating the length

these two don't go well together. If the value itself isn't length-describing,
like ASN.1 things, then the field should be shown as a separate field in the
struct? Especially since issuer_name claims its minimum length is 1 and so it
cannot contain the 2 octet length field. If the 2^16-1 implies the two octet
length, then perhaps a note is useful to state if this field is omitted, two
zero bytes are present? That is also the only way one could distinguish the
two optional fields redemption_context and origin_info if only one was included?
(this might be obvious to TLS13 readers and its syntax, but as specified, I
don't find it very clear, and I suspect non-TLS implementers might not either)

### LINKED ISSUE #384: origin_info clarifications
From @paulwouters:

On origin_info:

>        If the string contains multiple origin names, they are delimited
>        with commas "," without any whitespace.

What is a "string" ? (the same question can be asked for issuer_name). Is it null terminated? ASCII? UTF-8? A-label FQDN?

Does origin_info support strings containing a comma? If so how is this expressed?
Using \, or ,, or ? Is it valid for origin names to have leading or trailing spaces?
This seems a bit fragile to me, and perhaps some additional text can make it less
fragile?

chris-wood:
Closes #383 
Closes #384 

We might want to be a bit more crisp about the type of a `hostname`. It's generally a domain name, and with and without a trailing dot is acceptable. I'm comfortable with the text as-is, but if folks feel strongly about adding more specificity, let's do it.

cc @paulwouters for 👀


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -187,13 +187,13 @@ lengths for this field are either 0 or 32 bytes. The field is prefixed with a
...
 of invalid lengths MUST be ignored.
 
-- "origin_info" is an optional string containing one or more origin names,
-which allows a token to be scoped to a specific set of origins. The string
-is prefixed with a 2-octet integer indicating the length, in network byte
-order. If empty, any non-origin-specific token can be redeemed. If the string
-contains multiple origin names, they are delimited with commas "," without any
-whitespace. If this field is not empty, the Origin MUST include its own name
-as one of the names in the list.
+- "origin_info" is an optional value containing one or more origin hostnames,
```
tfpauly, line 190: Same for calling this out as UTF-8?

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -187,13 +187,13 @@ lengths for this field are either 0 or 32 bytes. The field is prefixed with a
...
+which allows a token to be scoped to a specific set of origins. The field
+that stores this value in the challenge is prefixed with a 2-octet integer
+indicating the length, in network byte order. If empty, any non-origin-specific
+token can be redeemed. If the value contains multiple origin hostnames, they are
+delimited with commas "," without any whitespace. If this field is not empty,
```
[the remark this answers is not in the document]
chris-wood: I figured that was implied from the fact that these are hostnames, but I'm happy to include it.

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -484,15 +484,15 @@ occurred in a given time window.
 
 Token challenges that include non-empty origin_info bind tokens to one or more
 specific origins. As described in {{challenge}}, clients only accept such
-challenges from origin names listed in the origin_info string. Even if multiple
+challenges from origin names listed in the origin_info value. Even if multiple
```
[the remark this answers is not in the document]
chris-wood: Reverted.

chris-wood:
@tfpauly fixed the changes. @paulwouters, can you please let us know if this resolves your two comments about `origin_info` and `issuer_name`?

tfpauly (approved):

draft-ietf-privacypass-auth-scheme.md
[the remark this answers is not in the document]
paulwouters: It is implied but it is always good to give a nudge to what could be a security issue if not considered, eg malicious hostnames with comma's

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -173,10 +173,10 @@ in an IANA registry, {{token-types}}. Challenges with unsupported token_type
 values MUST be ignored. This value determines the structure and semantics of
 the rest of the structure.
 
-- "issuer_name" is a string containing the name of the issuer. This is a
-hostname that is used to identify the issuer that is allowed to issue
-tokens that can be redeemed by this origin. The string is prefixed with a
-2-octet integer indicating the length, in network byte order.
+- "issuer_name" is a UTF-8 string containing the hostname of the issuer.
```
paulwouters, line 176: Is it UTF-8 hostnames? eg Ulabel? not A-label (eg punycode) ?

If the unit these things operate on are "dns hostnames", then I would expect this to use A-labels. That is not nohâts.ca but xn--nohts-5qa.ca

chris-wood: It's just a FQDN without the trailing dot, e.g., "github.com" or "ietf.org". Basically, the same thing that would go in the TLS SNI. It's _not_ an A-label with the bizarre xn-- prefix. I'll try to clarify.

chris-wood: @paulwouters how does the new text look?

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -187,13 +187,14 @@ lengths for this field are either 0 or 32 bytes. The field is prefixed with a
...
-order. If empty, any non-origin-specific token can be redeemed. If the string
-contains multiple origin names, they are delimited with commas "," without any
-whitespace. If this field is not empty, the Origin MUST include its own name
-as one of the names in the list.
+- "origin_info" is an optional UTF-8 string containing one or more origin hostnames,
+which allows a token to be scoped to a specific set of origins. The field
+that stores this string in the challenge is prefixed with a 2-octet integer
+indicating the length, in network byte order. If empty, any non-origin-specific
+token can be redeemed. If the string contains multiple origin hostnames, they are
+delimited with commas "," without any whitespace. Each origin hostname MUST NOT
```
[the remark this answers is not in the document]
chris-wood: Fixed.

draft-ietf-privacypass-auth-scheme.md
tfpauly: Yeah, this is the HTTP / URI style version of the hostname, not the DNS version of the name.

paulwouters:
On Jun 21, 2023, at 16:56, Tommy Pauly ***@***.***> wrote:﻿
Yeah, this is the HTTP / URI style version of the hostname, not the DNS version of the name.You mean what I type in my address bar ? I think when I type nohâts.ca  I think it will do an a-label lookup ? Anyway, make it very clear.

tfpauly:
Yes, the thing in the address bar

chris-wood:
> Yes, the thing in the address bar

I don't want to continue adding to the confusion here, but just so we're on the same page: I would describe this as the same exact format as the TLS SNI field. I don't _think_ that's the same as the thing as what one would punch into the address bar, since the latter allows all sorts of character sets?

paulwouters:
```
"issuer_name" is a UTF-8 string containing the hostname of the issuer, i.e., the
fully qualified DNS hostname of the issuer using ASCII encoding without a trailing dot.
```

That's a weird way of saying "ASCII".

tfpauly:
The SNI is defined as using A-labels:

> "HostName" contains the fully qualified DNS hostname of the server,
>    as understood by the client.  The hostname is represented as a byte
>    string using ASCII encoding without a trailing dot.  This allows the
>    support of internationalized domain names through the use of A-labels
>    defined in [[RFC5890](https://datatracker.ietf.org/doc/html/rfc5890)].  DNS hostnames are case-insensitive.  The
>    algorithm to compare hostnames is described in [[RFC5890](https://datatracker.ietf.org/doc/html/rfc5890)], [Section](https://datatracker.ietf.org/doc/html/rfc6066#section-2.3.2.4)
>    [2.3.2.4](https://datatracker.ietf.org/doc/html/rfc6066#section-2.3.2.4).

The other approach is to have it be what HTTP uses for authority, which I think makes sense given the layer we are operating at:

> ":authority":
> Contains the authority portion of the target URI ([Section 3.2](https://www.rfc-editor.org/rfc/rfc3986.html#section-3.2) of [[URI]](https://httpwg.org/specs/rfc9114.html#URI)). The authority MUST NOT include the deprecated userinfo subcomponent for URIs of scheme "http" or "https".

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -173,10 +173,12 @@ in an IANA registry, {{token-types}}. Challenges with unsupported token_type
...
-- "issuer_name" is a string containing the name of the issuer. This is a
-hostname that is used to identify the issuer that is allowed to issue
-tokens that can be redeemed by this origin. The string is prefixed with a
-2-octet integer indicating the length, in network byte order.
+- "issuer_name" is a UTF-8 string containing the hostname of the issuer, i.e., the
+fully qualified DNS hostname of the issuer using ASCII encoding without a trailing dot.
+This allows the support of internationalized hostnames through the use of A-labels defined in {{?RFC5890}}.
+This is hostname identifies the issuer that is allowed to issue tokens that can be
+redeemed by this origin. The field that stores this string in the challenge
+is prefixed with a 2-octet integer indicating the length, in network byte order.
```
tfpauly, line 181:
```suggestion
- "issuer_name" is an ASCII string that identifies the issuer using the format of the authority portion of a URI
as defined in {{Section 3.2 of !URI=RFC3986}}. This name identifies the issuer that is allowed to
issue tokens that can be redeemed by this origin. The field that stores this string in the challenge
is prefixed with a 2-octet integer indicating the length, in network byte order.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -187,13 +189,17 @@ lengths for this field are either 0 or 32 bytes. The field is prefixed with a
...
+each of which is a fully qualified DNS hostname using ASCII encoding without a trailing dot.
+This allows the support of internationalized hostnames through the use of A-labels defined in {{?RFC5890}}.
+The purpose of this field is to allow a token to be scoped to a specific set of origins.
+The field that stores this string in the challenge is prefixed with a 2-octet integer
+indicating the length, in network byte order. If empty, any non-origin-specific
+token can be redeemed. If the string contains multiple origin hostnames, they are
+delimited with commas "," without any whitespace. Moreover, any whitespace in the
+string MUST be ignored. Each origin hostname MUST NOT include a comma "," element.
+If the "origin_info" field is not empty, the Origin MUST include its own name as
+one of the hostnames in the list.
```
tfpauly, line 202:
```suggestion
- "origin_info" is an optional ASCII string containing one or more origin names, each of which
uses the format of the authority portion of a URI as defined in {{Section 3.2 of URI}}. 
The purpose of this field is to allow a token to be scoped to a specific set of origins.
The field that stores this string in the challenge is prefixed with a 2-octet integer
indicating the length, in network byte order. If empty, any non-origin-specific
token can be redeemed. If the string contains multiple origin hostnames, they are
delimited with commas "," without any whitespace. Moreover, any whitespace in the
string MUST be ignored. Each origin hostname MUST NOT include a comma "," character.
If the "origin_info" field is not empty, the Origin MUST include its own name as
one of the hostnames in the list.
```

tfpauly (commented):
Suggesting to use URI format for encoding, since we're in HTTP and that's what the application is already dealing with.

chris-wood:
The existing text pulls directly from RFC 6066, so I think we ought to keep it.

tfpauly:
@paulwouters please check the text now, should be clear what the format is

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -187,13 +187,16 @@ lengths for this field are either 0 or 32 bytes. The field is prefixed with a
...
 
-- "origin_info" is an optional string containing one or more origin names,
-which allows a token to be scoped to a specific set of origins. The string
-is prefixed with a 2-octet integer indicating the length, in network byte
-order. If empty, any non-origin-specific token can be redeemed. If the string
-contains multiple origin names, they are delimited with commas "," without any
-whitespace. If this field is not empty, the Origin MUST include its own name
-as one of the names in the list.
+- "origin_info" is an optional ASCII string containing one or more origin names, each of which
+uses the format of the authority portion of a URI as defined in {{Section 3.2 of URI}}.
```
tfpauly, line 191:
```suggestion
uses the format of the authority portion of a URI as defined in {{Section 3.2 of URI}}.
```

PR #391, opened 2023-06-21
Tokens are one-time-use

chris-wood:
We never actually say this explicitly, so I figured it'd be good to say it.


tfpauly:
I don't know if the one-time-use is necessarily a fundamental property of the auth scheme — it can be for a particular issuance protocol, but a new token type could theoretically have reused tokens for some use case. (I don't think that's generally a good idea, but I'm mainly saying this doesn't belong here)

chris-wood:
Hmm, I don't agree with that. The text already hinges on the need to implement some form of double spending prevention at the origins to prevent replays, which means that replays (using more than once) is not intended. The document seems like it has a very clear stance on this.

tfpauly:
Yeah, it does cover double spending as a SHOULD today. But it does leave a door open to other behaviors that still work with the auth scheme. I know your change isn't normative, but I see the one-time-use as a SHOULD, but not necessarily a MUST.

chris-wood:
I don't think we want to encourage those use cases. If they do arise, they can update this document to relax the requirement.

tfpauly:
Okey doke

tfpauly (approved):

PR #392, opened 2023-06-22
Clarify default structures

### LINKED ISSUE #382: token_type scope clarification
From @paulwouters

        "token_type" is a 2-octet integer, in network byte order. This
        type indicates the issuance protocol used to generate the
        token. Values are registered in an IANA registry, Section
        5.2. Challenges with unsupported token_type values MUST be
        ignored. This value determines the structure and semantics of
        the rest of the structure.
        [ more fields of the structure ]

If now describing more structures, those can only be valid for a specific
value of token_type. This should then be mentioned here. (same issue in
"Token Redemption" section). Also, I don't see any non-grease value added
to the IANA registry. I assume 0x0002 will be added for [ISSUANCE] and
that the structures here are for that use case? If so, that deserves
clarification. If not, I think this also might need some rephrasing?

tfpauly:
Closes #382


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -167,29 +184,24 @@ struct {
...
-to require that clients fetch tokens bound to a specific context, as opposed
-to reusing tokens that were fetched for other contexts. See
+- "redemption_context" is an optionally empty field. If non-empty, it allows
+the origin to require that clients fetch tokens bound to a specific context, as
+opposed to reusing tokens that were fetched for other contexts. See
 {{context-construction}} for example contexts that might be useful in practice.
 When present, this value is a 32-byte context generated by the origin. Valid
 lengths for this field are either 0 or 32 bytes. The field is prefixed with a
 single octet indicating the length. Challenges with redemption_context values
 of invalid lengths MUST be ignored.
```
tfpauly, line 202:
```suggestion
- "redemption_context" is a field that is either 0 or 32 bytes, prefixed with a single
octet indicating the length (either 0 or 32). If value is non-empty, it is a 32-byte value
generated by the origin that allows the origin to require that clients fetch tokens
bound to a specific context, as opposed to reusing tokens that were fetched for other
contexts. See {{context-construction}} for example contexts that might be useful in
practice. Challenges with redemption_context values of invalid lengths MUST be ignored.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -167,29 +184,24 @@ struct {
...
 
-- "origin_info" is an optional string containing one or more origin names,
+- "origin_info" is an optionally empty string containing one or more origin names,
 which allows a token to be scoped to a specific set of origins. The string
 is prefixed with a 2-octet integer indicating the length, in network byte
 order. If empty, any non-origin-specific token can be redeemed. If the string
```
tfpauly, line 207:
```suggestion
- "origin_info" is a string that is either empty, or contains one or more origin
names that allow a token to be scoped to a specific set of origins. The string
is prefixed with a 2-octet integer indicating the length, in network byte
order. If empty, any non-origin-specific token can be redeemed. If the string
```

paulwouters:
Thanks. I reviewed the text, and it is okay now.

PR #410, opened 2023-07-12
Add token  key encoding to the registry

### LINKED ISSUE #409: Token key encoding ambiguity
There is currently no text that says precisely how token keys are encoded. As an example, one could infer that it's as a DER encoded SPKI blob based on how the token key ID is computed, but we're not explicit about that. We should fix that.

chris-wood:
Closes #409

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -753,7 +753,7 @@ This document updates the "Token Type" Registry from
 * Value: 0x0002
 * Name: Blind RSA (2048-bit)
 * Token Structure: As defined in {{Section 2.2 of AUTHSCHEME}}
-* Token Key Encoding: Serialized as a DER-encoded SubjectPublicKeyInfo (SPKI) object
+* Token Key Encoding: Serialized as a DER-encoded SubjectPublicKeyInfo (SPKI) object, using the RSASSA-PSS OID {{!RFC5756}}
 * TokenChallenge Structure: As defined in {{Section 2.1 of AUTHSCHEME}}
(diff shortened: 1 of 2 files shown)


draft-ietf-privacypass-protocol.md
```diff
@@ -752,6 +753,7 @@ This document updates the "Token Type" Registry from
 * Value: 0x0002
 * Name: Blind RSA (2048-bit)
 * Token Structure: As defined in {{Section 2.2 of AUTHSCHEME}}
+* Token Key Encoding: Serialized as a DER-encoded SubjectPublicKeyInfo (SPKI) object
```
thibmeu, line 756: maybe mention [RFC2692](https://www.rfc-editor.org/rfc/rfc2692.html)

chris-wood: I went with a variation 👍

draft-ietf-privacypass-protocol.md
```diff
@@ -752,6 +753,7 @@ This document updates the "Token Type" Registry from
 * Value: 0x0002
 * Name: Blind RSA (2048-bit)
 * Token Structure: As defined in {{Section 2.2 of AUTHSCHEME}}
+* Token Key Encoding: Serialized as a DER-encoded SubjectPublicKeyInfo (SPKI) object, using the RSASSA-PSS OID {{!RFC5756}}
```
chris-wood, line 756:
```suggestion
* Token Key Encoding: Serialized as a DER-encoded SubjectPublicKeyInfo (SPKI) object using the RSASSA-PSS OID {{!RFC5756}}
```

PR #413, opened 2023-07-28
Revise padding rationale

### LINKED ISSUE #412: Auth scheme: update rationale for base64 padding guidance
In the [current auth scheme draft](https://www.ietf.org/archive/id/draft-ietf-privacypass-auth-scheme-11.html#section-2.1), `challenge` and `token-key` are currently defined as base64url-encoded values. The draft also states:
> Since the length of the challenge is not fixed, the base64url value MUST include padding.

This sounds like a misunderstanding of how base64 padding works. Padding is only required when concatenating two base64 messages. Here's an example:

* `hello` ASCII encodes as base64 `aGVsbG8=`
* `base` ASCII encodes as base64 `YmFzZQ==`

On the decoding side, `aGVsbG8` is non-ambiguous: it decodes just as well as `aGVsbG8=`. Similarly, if you see `YmFzZQ` you know how many bytes there are and it decodes fine. The problem arises if you concatenate: `aGVsbG8=YmFzZQ==` decodes to `hellobase` (assuming your decoder supports this which many don't), but `aGVsbG8YmFzZQ` decodes to `hello\`.

In the context of this draft, the message is strongly delimited (ends with the end of the header, or with a whitespace or a comma etc.) so there is no opportunity for concatenation.

This document shouldn't be requiring padding.

tfpauly:
Closes #412

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -212,18 +212,18 @@ When used in an authentication challenge, the "PrivateToken" scheme uses the
 following parameters:
 
 - "challenge", which contains a base64url-encoded {{!RFC4648}} TokenChallenge
- value. Since the length of the challenge is not fixed, the base64url value
- MUST include padding. As an Authentication Parameter (`auth-param` from
- {{!RFC9110, Section 11.2}}), the value can be either a token or a
- quoted-string, and might be required to be a quoted-string if the base64url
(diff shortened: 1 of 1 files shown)


chris-wood (approved):
Much better! Did we ever sort out where the current text rationale came from?

tfpauly:
The old rationale was back from before this document was adopted, as part of the original individual document.

PR #424, opened 2023-08-03
Clearly specify token verification steps

### LINKED ISSUE #415: DISCUSS: under-specification
(a) There is guidance in the architecture and issuance documents; and this
document to construct an end-to-end solution.  However, for the purposes of
these document, those other are only informative.  There appears to be a few
places of under-specification or implicit assumptions.

** Section 2.2
   For token types that support public verifiability, origins verify the
   token authenticator using the public key of the issuer, and validate
   that the signed message matches the concatenation of the client nonce
   and the hash of a valid TokenChallenge.

-- Please explain what “public verifiability” means.  I didn’t see this term in
the architecture document.

-- Implementation details of the authenticator/token seem to be leaking into
this text (i.e., properties of the nonce || hash TokenChallenge).  Does this
suggest requirements for the construction of the token? Put in another way,
where is the normative guidance that requires that construction?  I couldn’t
find other language in this document on the cryptographic properties of the
Token.

** Section 2.2.  Since this section is describing the redemption process, I
missed something obvious -- how does the origin know it got a “good” token”.  I
was expecting to see language which say that there is a token-specific
verification steps of the Token’s cryptographic assurances.

### LINKED ISSUE #420: COMMENT: Authenticator construction
** Section 2.2.
*  "authenticator" is a Nk-octet authenticator that covers the
      preceding fields in the token.  The value of this field is defined
      by the token_type and corresponding issuance protocol.  The value
      of constant Nk depends on token_type, as defined in Section 5.2.

-- Could a bit more be said about the purpose of this field.  Please also
explain what “cover” means in this context.

-- Will the security properties of these authenticators vary by token_type?  If
so, please explicitly state that?

chris-wood:
This references the issuance protocol document for details.

Closes #420
Closes #415

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -407,7 +407,7 @@ Authorization: PrivateToken token="abc..."
 ~~~
 
 For context-bound tokens, origins store or reconstruct the contexts of previous
-TokenChallenge structures in order to validate the token. A TokenChallenge MAY
+TokenChallenge structures in order to validate the token. A TokenChallenge can
 be bound to a specific TLS session with a client, but origins can also accept
 tokens for valid challenges in new sessions. Origins SHOULD implement some form
 of double-spend prevention that prevents a token with the same nonce from being
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -379,10 +379,11 @@ original TokenChallenge, SHA256(TokenChallenge).
 key. The value of this field is defined by the token_type and corresponding
 issuance protocol.
 
-- "authenticator" is a Nk-octet authenticator that covers the preceding fields
-in the token. The value of this field is defined by the token_type and
-corresponding issuance protocol. The value of constant Nk depends on
-token_type, as defined in {{token-types}}.
+- "authenticator" is a Nk-octet authenticator that is cryptographically bound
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -403,22 +406,45 @@ the Authorization header field as follows:
...
-the contexts of previous TokenChallenge structures in order to validate the
-token. A TokenChallenge MAY be bound to a specific TLS session with a client,
-but origins can also accept tokens for valid challenges in new sessions.
-Origins SHOULD implement some form of double-spend prevention that prevents
-a token with the same nonce from being redeemed twice. This prevents clients
-from "replaying" tokens for previous challenges. For context-bound tokens,
-this double-spend prevention can require no state or minimal state, since
-the context can be used to verify token uniqueness.
+For context-bound tokens, origins store or reconstruct the contexts of previous
+TokenChallenge structures in order to validate the token. A TokenChallenge MAY
```
tfpauly, line 410: Now that we're going back to this, does this need to be a normative MAY, or can it be:
```suggestion
TokenChallenge structures in order to validate the token. A TokenChallenge can
```

tfpauly (approved):

draft-ietf-privacypass-auth-scheme.md
chris-wood: This was existing text, so I'm fine relaxing it.

tfpauly (approved):

PR #425, opened 2023-08-03
Clean up authn and authz relationship in the intro

### LINKED ISSUE #395: Artart review
# Major Concerns

In Section 1, I think that a bit more context is necessary.  At a very high level, this is an architecture for authorization based on privacy-preserving authentication mechanisms.  So, this section needs to say more about authentication, authorization, and how the two work together.  See the discussion of these two concepts in RFC 4949.

In Section 2, I think that the definition of Attestation procedure should be reworded to avoid using "trusted" in this manner.  The use of "trust" In Section 3.2 is more in line with the usual use of this term.  Suggestion:

```
   Attestation procedure:  The procedure by which an Attester determines
      whether or not a Client has the specific set of properties 
      that are necessary for token issuance.
```

# Minor Concerns

Section 5: Please explain "viability of an open Web".

# Nits

Section 3.4.2: A reference for Tor would be useful.

chris-wood:
Closes #395


draft-ietf-privacypass-architecture.md
```diff
@@ -84,35 +84,45 @@ informative:
...
-or not the client is an authorized user or has completed some prior
-challenge, clients present unlinkable proofs that attest to this information.
-These proofs, or tokens, are private in the sense that a given token cannot
-be linked to the protocol interaction where that token was initially issued.
+authentication mechanisms. In other words, relying parties authenticate clients
+in a privacy-preserving way, i.e., without learning any unique, per-client
+information through the authentication protocol, and then make authorization
+decisions on the basis of that authentication suceeding or failing. Possible
+authorization decisions might be to provide clients with read access to a
+particular resource or write access to a particular API.
```
tfpauly, line 103: I find the use of "API" here to be a bit odd. I would lean towards implying this and just saying that the authorization allows clients to access a resource.

draft-ietf-privacypass-architecture.md
```diff
@@ -856,23 +869,26 @@ Origin from a token request.
...
-of Clients based on Attestation support. This could lead to harmful ecosystem
-effects if left unresolved.
-
-In principle, Issuers should strive to work with a set of Attesters that are
-suitable for all Clients, thereby mitigating such discriminatory behavior.
-In practice, this may require tradeoffs in what type of attestation Issuers are
-willing to trust so as to enable more widespread support.
+of Clients based on Attestation support. For example, an Origin could only
+authorize Clients that successfully authenticate with a token, prohibiting access
+to all other Clients. If the Attestation is not equitable, this can yield
```
tfpauly, line 884:
```suggestion
to all other Clients. If the Attestation is not equally available to all legitimate clients, this can yield
```

chris-wood: I don't think we can say this without also defining what legitimate means, and I don't think we have a good definition. That's why I went with equitable.

draft-ietf-privacypass-architecture.md
chris-wood: Works for me!

chris-wood, line 103:
```suggestion
particular resource or write access to a particular resource.
```

draft-ietf-privacypass-architecture.md
tfpauly: I feel that "equitable" is not necessarily clearly defined here either. That seems to imply some broader statement, as opposed to just saying that "anyone who should be able to pass the bar can". If the use for a particular token is saying "you have a legitimate account to use this privacy proxy or OHTTP relay", then it's OK for that to be limited. If the use is to access a website, it needs to be open to all browsers. Etc, etc.

chris-wood: I agree, neither are really clearly defined. What if we just gave examples (like those which you listed)?

tfpauly: Sure, maybe giving more concrete examples is the best route here.

chris-wood: @tfpauly please take a look at the latest version and let me know what you think?

tfpauly (approved):

PR #431, opened 2023-08-07
No provisional registry use

### LINKED ISSUE #430: Registry guidance
### Section 5.2, paragraph 15
```
     This registry also will also allow provisional registrations to allow
     for experimentation with protocols being developed.  Designated
     experts review, approve, and revoke provisional registrations.
```
How/why/when do experts revoke provisional registrations? AFAIK this
is not something that they usually do for other registries. Can
provisional registrations be upgraded to regular ones without
codepoint changes? This all might need more text, or maybe we can
rely on the "private use" range for this?

chris-wood:
Closes #430

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -592,12 +592,9 @@ ensure that the token type is sufficiently clearly defined to be used for both
 token issuance and redemption, and meets the common security and privacy
 requirements for issuance protocols defined in {{Section 3.2 of ARCHITECTURE}}.
 
-This registry also will also allow provisional registrations to allow for
-experimentation with protocols being developed. Designated experts review,
-approve, and revoke provisional registrations.
-
-Values 0xFF00-0xFFFF are reserved for private use, to enable proprietary uses
-and limited experimentation.
+Values 0xFF00-0xFFFF are reserved for private use. Implementers can use values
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -592,12 +592,9 @@ ensure that the token type is sufficiently clearly defined to be used for both
...
 
-This registry also will also allow provisional registrations to allow for
-experimentation with protocols being developed. Designated experts review,
-approve, and revoke provisional registrations.
-
-Values 0xFF00-0xFFFF are reserved for private use, to enable proprietary uses
-and limited experimentation.
+Values 0xFF00-0xFFFF are reserved for private use. Implementers can use values
+in this range for experimentation with new token type protocols, as well as other
+proprietary uses that do not require interoperability.
```
dvorak42, line 597: Wonder if we should have an extra line that implementers should not rely on this being compatible on the public internet (since other private use implementations may collide). Though might be implicit in the "interoperability" bit.

dvorak42 (approved):

draft-ietf-privacypass-auth-scheme.md
chris-wood: Right, that's the intent of that sentence.

tfpauly (approved):

PR #432, opened 2023-08-07
SECDIR review

### LINKED ISSUE #397: [Secdir review] Terminology
```
   The following terms are used throughout this document:

   Client:  An entity that seeks authorization to an Origin.
```

- [nms] It would be helpful to cite RFC9334 sooner than midway through section 3 and add a note to this definition that "Client implements the RATS Attester  role." Additionally, it isn't clear if "entity" refers to the RFC4949 definition or is more constrained. This is important when considering privacy protections as protection of user (human) trackability isn't necessarily 1 to 1 with device or platform trackability. 

```
   Origin:  An entity that redeems tokens presented by Clients.
```
- [nms]  "token redemption" is not a well-known phrase that warrant explanation. Normally, privacy preservation is an objective of the Client. The other entities in the system are potential adversaries. This definition doesn't specify whether the Origin is trusted by the Client or a privacy adversary.

```
   Issuer:  An entity that issues tokens to Clients for properties attested to by the Attester.
```
- [nms]  This definition doesn't specify whether the Issuer is trusted by the Client or a privacy adversary.

```
   Attester:  An entity that attests to properties of Clients for the purposes of token issuance.
```
- [nms]  This definition doesn't specify whether the Issuer is trusted by the Client or a privacy adversary.
- [nms]  I suggest using "Attestation Verifier" so as not to create reader dissonance for those familiar with RFC9334. 

```
   Attestation procedure:  The procedure by which an Attester determines whether or not a Client is trusted with a specific set of properties for token issuance.
```
- [nms]  This definition isn't clear whether "trusted with a specific set of properties" means the Client is authorized to possess the set of properties or whether the Client possessing the properties may be the wrong set of properties for one of the other entities to trust the Client.

### LINKED ISSUE #398: [Secdir review] Use cases
- [nms]  There should be a use cases section or reference to stand-alone use cases document that explains the context for who might deploy a Privacy Pass solution and the conditions in which such deployment is necessary.

### LINKED ISSUE #399: [Secdir review] Architecture overview and example
```
3.  Architecture

   The Privacy Pass architecture consists of four logical entities --
   Client, Origin, Issuer, and Attester -- that work in concert for
   token redemption and issuance.  This section presents an overview of
   Privacy Pass, a high-level description of the threat model and
   privacy goals of the architecture, and the goals and requirements of
   the redemption and issuance protocols.
```
- [nms]  This definition doesn't describe the relevance of token issuance (or refusal to issue) as a mechanism for privacy protection (preservation). A primary concern for user privacy is that attributes that can be regarded as PII or PI can be found (inferred through differential privacy analysis) from a collection of other attributes that separately are not classified as PI or PII but together reveal a value that should be classified as PI or PII. It isn't clear if the token is intended as a set of attributes that survives DPA, a handle that identifies the user, but can't be compromised using DPA, something else?

```
3.1.  Overview

   The typical interaction flow for Privacy Pass uses the following
   steps:

   1.  A Client interacts with an Origin by sending a request or
       otherwise interacting with the Origin in a way that triggers a
       response containing a token challenge.  The token challenge
       indicates a specific Issuer to use.
```
- [nms]  What are some examples of triggering mechanisms or circumstances? 

```
   2.  If the Client already has a token available that satisfies the
       token challenge, e.g., because the Client has a cache of
       previously issued tokens, it can skip to step 6 and redeem its
       token; see Section 7.1 for security considerations of cached
       tokens.
```
- [nms]  Step 1 says the result of step 1 is an indication of which Issuer to use, not that a token was issued. Hence, the cached information from step 1 is a suggested token issuer (not an issued token).

```
   3.  If the Client does not have a token available and decides it
       wants to obtain one (or more) bound to the token challenge, it
       then invokes the issuance protocol.  As a prerequisite to the
       issuance protocol, the Client runs the deployment specific
       attestation process that is required for the designated Issuer.
       Client attestation can be done via proof of solving a CAPTCHA,
```

- [nms]  "Attestation" via CAPTCHA is a proof of liveness (rather than an attestation of device trustworthiness properties). The sentence structure seems to conflate these two very different properties.

```
   4.  If the attestation process completes successfully, the client
       creates a Token Request to send to the designated Issuer
       (generally via the Attester, though it is not required to be sent
       through the Attester).  
```

- [nms]  Wouldn't the token request be made prior to the interaction with the Attester where the interaction is a way to populate one of the fields in the token request? Using RATS passport model, the token request contains the attestation passport that is then sent to the Issuer. Alternatively, the Token Request is sent to the Attester who forwards it to the Issuer who then does a RATS background check prior to issuing the token response. 

```
The Attester and Issuer might be
       functions on the same server, depending on the deployment model
       (see Section 4).  Depending on the attestation process, it is
       possible for attestation to run alongside the issuance protocol,
       e.g., where Clients send necessary attestation information to the
       Attester along with their Token Request.  If the attestation
       process fails, the Client receives an error and issuance aborts
       without a token.

   5.  The Issuer generates a Token Response based on the Token Request,
       which is returned to the Client (generally via the Attester).
       Upon receiving the Token Response, the Client computes a token
       from the token challenge and Token Response.  This token can be
       validated by anyone with the per-Issuer key, but cannot be linked
       to the content of the Token Request or Token Response.

   6.  If the Client has a token, it includes it in a subsequent request
       to the Origin, as authorization.  

```
- [nms]  Normally, a server that requires authorization might be using OAuth2. Does Privacy Pass integrate with OAuth2 or if not, does it compete with OAuth2?

```
This token is sent only once.
```

- [nms]  Does this mean it is only sent once per Origin (i.e., each Origin receives a different token) or that for a given Origin, it receives a token one time? Given the Origin-Client interaction is RESTful, and there are multiple REST interactions, will the Client obtain a different token for each RESTful exchange?

```
       The Origin validates that the token was generated by the expected
       Issuer and has not already been redeemed for the corresponding
       token challenge.  If the Client does not have a token, perhaps
       because issuance failed, the client does not reply to the
       Origin's challenge with a new request.
```
- [nms]  It isn't clear if "Request + Token" is combing two otherwise independent flows: (a) Token - which completes the flow described in Fig. 1. and (b) Request - which restarts the flow in Fig. 1. Or if Request+Token finalizes the flow in Fig. 1. If so, what is being requested in "Request"?

### LINKED ISSUE #400: [Secdir] Privacy goals and threat model
General comment:

- [nms] The threat model should also describe the obligations each entity is under for not disclosing attributes obtained from the Client. If attributes are shared, what obligations the targets of sharing are under to not disclose etc. Given the Issuer and Origin(s) are adversaries of the Client and given the Attester has a dual role of being both a trusted entity of the Client and a trusted entity of the Issuer, the potential for Attester being manipulated into bias against the Client seems likely. The threat model should do more to explain how these (contradictory) objective can be met. 

The unlinkability goals are necessarily overridden by the attestation objective to share Client attributes. Hence more should be said about how the Client is supposed to trust its adversaries with information that could be shared with colluding partners.

```
   The privacy goals of Privacy Pass assume a threat model in which
   Origins trust specific Issuers to produce tokens, and Issuers in turn
   trust one or more Attesters to correctly run the attestation
   procedure with Clients.  This arrangement ensures that tokens which
   validate for a given Issuer were only issued to a Client that
   successfully completed attestation with an Attester that the Issuer
   trusts.  Moreover, this arrangement means that if an Origin accepts
   tokens issued by an Issuer that trusts multiple Attesters, then a
   Client can use any one of these Attesters to issue and redeem tokens
   for the Origin.
```
- [nms]The threat model should consider the possibility that any of the entities could collude with any other entities to obtain Redemption context, Issuance context, and Attestation context. Based on the descriptions above, these sets are not disjoint, hence, portions of the set of attributes associated with the Client are shared by some of the entities. Since, the Privacy Pass scope doesn't attempt to understand the various attestation protocols (and therefore doesn't understand which attributes are used), it doesn't seem reasonable to assume that there is any difference in terms of the attributes of each of the contexts. Consequently, (through collusion or otherwise) all entities have access to all attributes, the Privacy Pass protocol would not provide an effective protection against DPA attacks vs. the Client withholding all attributes that contribute to DPA attack weakness.

```
   Clients explicitly trust Attesters to perform attestation correctly
   and in a way that does not violate their privacy.  
```

- [nms]This type of trust should be explained more clearly as RFC9334 doesn't claim to perform attestation that is privacy preserving. Except for draft-ietf-rats-daa-03 which focuses only on privacy properties of RATS attester signing keys. How does the Client evaluate an Attester (RATS Verifier) to assess whether it is capable of preserving Client privacy while also fulfilling the requirements of Relying Parties desire to gain insight about Client (RATS Attester) attributes.

```
However, Clients assume Issuers and Origins are malicious.
```
- [nms]It would be helpful to include this assumption as part of the description of the roles (and reference this section for details).

```
   Given this threat model, the privacy goals of Privacy Pass are
   oriented around unlinkability based on redemption, issuance, and
   attestation contexts, as described below.

   1.  Origin-Client unlinkability.  This means that given two
       redemption contexts, the Origin cannot determine if both
       redemption contexts correspond to the same Client or two
       different Clients.  Informally, this means that a Client in a
       redemption context is indistinguishable from any other Client
       that might use the same redemption context.  The set of Clients
       that share the same redemption context is referred to as a
       redemption anonymity set.
```
- [nms]  Based on previous description, it seems the anonymity sets may not be disjoint. If not, then understanding the DPA impact to anonymity cannot be strictly understood for a given anonymity set.

```
   2.  Issuer-Client unlinkability.  This is similar to Origin-Client
       unlinkability in that a Client in an issuance context is
       indistinguishable from any other Client that might use the same
       issuance context.  The set of Clients that share the same
       issuance context is referred to as an issuance anonymity set.

   3.  Attester-Origin unlinkability.  This is similar to Origin-Client
       and Issuer-Client unlinkability.  It means that given two
       attestation contexts, the Attester cannot determine if both
       contexts correspond to the same Origin or two different Origins.
       The set of Origins that share the same attestation context is
       referred to as an attestation anonymity set.

   4.  Redemption context unlinkability.  Given two redemption contexts,
       the Origin cannot determine which issuance and attestation
       contexts each redemption corresponds to, even with the
       collaboration of the Issuer and Attester (should these be
       different parties).  This means that any information that may be
       learned about the Client during the issuance and attestation
       flows cannot be used by the Origin to compromise Client privacy.

   By ensuring that different contexts cannot be linked in this way,
   only the Client is able to correlate information that might be used
```
- [nms]  this conclusion doesn't follow based on the potential for collusion and attribute overlap in the various anonymity sets.

### LINKED ISSUE #401: [Secdir review] Redemption
```
   *  Issuer identity.  Token challenges identify which Issuers are
      trusted for a given issuance protocol.  As described in
      Section 3.2, the choice of Issuer determines the type of Attesters
      and attestation procedures possible for a token from the specified
      Issuer, but the Client does not learn exactly which Attester was
      used for a given token issuance event.
```
- [nms] Earlier descriptions indicate the Issuer issues token request responses while the Attester provides results to the Client and the Client creates the token. How is it the Client doesn't know the Attester?

```
   *  Redemption context.  Challenges can be bound to a given redemption
```
- [nms] As an architecture document, does it make sense to describe the requirements of the system that achieve the privacy goals vs. describing the behavior that may or may not be privacy relevant?

```
      context, which influences a client's ability to pre-fetch and
      cache tokens.  For example, an empty redemption context always
      allows tokens to be issued and redeemed non-interactively, whereas
      a fresh and random redemption context means that the redeemed
      token must be issued only after the client receives the challenge.
      See Section 2.1.1 of [AUTHSCHEME] for more details.

   *  Per-Origin or cross-Origin.  Challenges can be constrained to the
```
- [nms] As an architecture document, does it make sense to describe the requirements of the system that achieve the privacy goals vs. describing the behavior that may or may not be privacy relevant?

```
      Origin for which the challenge originated (referred to as per-
      Origin tokens), or can be used across multiple Origins (referred
      to as cross-Origin tokens).  The set of Origins for which a cross-
      Origin token is applicable is referred to as the cross-Origin set.

   Origins that admit cross-Origin tokens bear some risk of allowing
   tokens issued for one Origin to be spent in an interaction with
   another Origin.  
```
- [nms] How is the risk quantified? 

```
In particular, depending on the use case, Origins
   may need to maintain state to track redeemed tokens.  For example,
   Origins that accept cross-Origin tokens across shared redemption
```
- [nms] A 'cross-origin shared redemption' sounds like collusion. What makes it not collusion from the perspective of the Client?

```
   contexts SHOULD track which tokens have been redeemed already in
   those redemption contexts, since these tokens can be issued and then
   spent multiple times in response to any such challenge.  
```
- [nms] Wouldn't the first "spend" of the token represent a disclosure of potentially privacy revealing activity? If so, wouldn't a subsequent "spend" equate to the same disclosure? Once information is disclosed the effect on privacy is idempotent. Assume that either the information is known everywhere or that its scope of disclosure is limited by the entity you disclosed it too, and that entity is trusted to maintain that scope regardless of the number of times it is re-disclosed.

### LINKED ISSUE #402: [Secdir review] Issuance
```
3.4.  Issuance Protocol

   The Privacy Pass issuance protocol, described in [ISSUANCE], is a
   two-message protocol that takes as input a TokenChallenge from the
   redemption protocol ([AUTHSCHEME], Section 2.1) and produces a Token
   ([AUTHSCHEME], Section 2.2), as shown in Figure 1.
```
- [nms] This architecture has several sections that are conceptually incomplete unless the reader reads the cited I-D that are WIP. This implies the architecture can't be completed until the dependent I-Ds are completed. Is that how the WG intends to move this I-D forward? Alternatively, the concepts referenced by the cited I-Ds can be included here and written in the style of requirements. The referenced I-Ds can be written in the style of an implementation that follows the requirements. References, become informative rather than normative. Otherwise, I assert these references are actually normative and are in the wrong section below.

```
3.4.1.  Attester Role

   In Privacy Pass, attestation is the process by which an Attester
   bears witness to, confirms, or authenticates a Client so as to verify
   properties about the Client that are required for Issuance.  Issuers
   trust Attesters to perform attestation correctly.
```
- [nms] A clearer definition of 'trust' is needed so that the reader can understand if the Issuer is a potential adversary that may 'leak' or expose information that can be used to perform DPA attacks or if the Issuer is itself doing DPA attacks. It isn't clear whether the Issuer is a RATS Relying Party or if Origin is a RATS Relying Party. If both are RATS RPs, then it isn't clear if the Client is trusting one more than the other to enforce the Client privacy policy.

```
   [RFC9334] describes an architecture for attestation procedures.
   Using that architecture as a conceptual basis, Clients are RATS
   attesters that produce attestation evidence, and Attesters are RATS
   verifiers that appraise the validity of attestation evidence.
```
- [nms] If Privacy Pass uses RFC9334 definitions of roles, then the previous paragraph is fallacious. A RATS Verifier do not attest, instead they produce attestation results while the Client attests.

```
   The type of attestation procedure is a deployment-specific option and
   outside the scope of the issuance protocol.  
```
- [nms] I don't understand how, if the algorithm that exposes Client attributes, is out of scope of the I-D, that Privacy Pass is able to understand the DPA properties of a particular disclosed attribute. This seems essential to any of the Privacy Pass entities that are trusted by the Client to enforce it's privacy policy.

```
   In general, each attestation procedure has different security
   properties.  For example, attesting to having a valid account is
   different from attesting to running on trusted hardware.  In general,
   minimizing the set of supported attestation procedures helps minimize
   the amount of information leaked through a token.
```
- [nms] Word choice is a concern as it suggests a Privacy Pass token can't be trusted to protect the privacy interests of the Client.

```
3.4.2.  Issuer Role

   In Privacy Pass, the Issuer is responsible for completing the
   issuance protocol for Clients that complete attestation through a
   trusted Attester.  As described in Section 3.4.1, Issuers explicitly
   trust Attesters to correctly and reliably perform attestation.
```
- [nms] Yes, but Client's trust Attesters not to disclose attributes that compromises Client privacy. The threat model said that Issuer and Origin were adversaries of the Client and the Attester was trusted by the Client (but didn't clearly explain which attributes the Client trusted the Attester not to disclose to the Issuer / Origin). How is does the Attester serve the interests of both Client and Issuer/Origin who are adversaries?

```
   Depending on the deployment model case, issuance may require some
   form of Client anonymization service,
```
- [nms] This seems to suggest there is a 5th role "Client Anonymization Service" that isn't mentioned in the threat model. Given the Issuer is a Client adversary, why should the Client trust the Client Anonymization Service which is deployed from the Issuer context? Wouldn't the Issuer already have access to Client attributes at this point? Given the Client already trusts the Attester to prevent disclosure of attributes to Issuer / Origin, why wouldn't the Client Anonymization Service be employed by the Client directly or by the Attester? 

```
 similar to an IP-hiding proxy,
```
- [nms] The IP-hiding proxy exists at a particular layer in the networking stack so as being in a position to prevent the next hop entity from observing the IP address. If this logic is applied to attributed "FOO", the application or network context is essential to determining if the Client Anonymization Service is an effective proxy. Yet there is not discussion about this important architectural dependency on a per attribute basis. Alternatively, the Client Anonymization Service is encrypting attributes to avoid disclosure. In this context, understanding the key exchange path is essential. 

```
   so that Issuers cannot learn information about Clients.  This can be
   provided by an explicit participant in the issuance protocol, or it
   can be provided via external means, such as through the use of an IP-
   hiding proxy service like Tor. In general, Clients SHOULD minimize or
   remove identifying information where possible when invoking the
   issuance protocol.
```
- [nms] Ultimately, what matters is whether the information contributes to a DPA attack. Unless there is a way for a Client to evaluate whether or not to "minimize or remove identifying information" comes down to whether the Client can assess whether the attribute contributes to a DPA attack. I don't see how that is enabled by Privacy Pass.

```
   Issuers are uniquely identifiable by all Clients with a consistent
   identifier.  In a web context, this identifier might be the Issuer
   host name.  
```
- [nms] Load balancers obscure the host name from the instance. Hence, host names are not necessarily unique identifiers of an entity. Issuers maintain one or more configurations, including

```
   Public metadata is that which clients can observe as part of the
   token issuance flow.  Public metadata can either be transparent or
   opaque.  For example, transparent public metadata is a value that the
   client either generates itself, or the Issuer provides during the
   issuance flow and the client can check for correctness.  Opaque
   public metadata is metadata the client can see but cannot check for
   correctness.  As an example, the opaque public metadata might be a
   "fraud detection signal", computed on behalf of the Issuer, during
   token issuance.  In normal circumstances, Clients cannot determine if
   this value is correct or otherwise a tracking vector.
```
- [nms] Is this saying the Client can't tell when it has been labeled with a "fraud detection signal"? Shouldn't the goals of Privacy Pass favor Client's ability to detect when a privacy disclosure event has occurred such that it can attempt to apply a privacy counter measure. Possibly, there are privacy laws that prevent misuse even if privacy sensitive information is disclosed.

### LINKED ISSUE #403: [Secdir review] Information flow
```
3.5.1.  Token Challenge Flow

   To use Privacy Pass, Origins choose an Issuer from which they are
   willing to accept tokens.  Origins then construct a token challenge
   using this specified Issuer and information from the redemption
   context it shares with the Client.  This token challenge is then
   delivered to a Client.  The token challenge conveys information about
   the Issuer and the redemption context, such as whether the Origin
   desires a per-Origin or cross-Origin token.  Any entity that sees the
   token challenge might learn things about the Client as known to the
   Origin.  
```
- [nms] The threat model didn't qualify the Origin vs. cross-Origin such that an Origin is trusted by the Client to protect the attributes disclosed to it from further disclosure to cross-Origin entities. 

```
3.5.2.  Attestation Flow

   Clients interact with the Attester to prove that they meet some
   required set of properties.  In doing so, Clients contribute
   information to the attestation context, which might include sensitive
   information such as application-layer identities, IP addresses, and
   so on.  Clients can choose whether or not to contribute this
   information based on local policy or preference.
```
- [nms] Policy or preference seems insufficient advice since the Client is motivated to participate in the online activity that the user is motivated to participate in. The privacy question is not easily all or nothing, but rather a question of whether disclosed attributes constitutes successful DPA attacks. 

```
   Exceptional cases in the issuance protocol, such as when either the
   Attester or Issuer aborts the protocol, can contribute information to
   the attestation or issuance contexts.  The extent to which
   information in this context harms the Issuer-Client or Attester-
   Origin unlinkability goals in Section 3.2 depends on deployment
   model; see Section 4.  Clients can choose whether or not to
   contribute information to these contexts based on local policy or
   preference.
```
- [nms] Policy or preference seems insufficient advice since the Client is motivated to participate in the online activity that the user is motivated to participate in. The privacy question is not easily all or nothing, but rather a question of whether disclosed attributes constitutes successful DPA attacks. 

```
3.5.4.  Token Redemption Flow

   Clients send tokens to Origins during the redemption protocol.  Any
   information that is added to the token during issuance can therefore
   be sent to the Origin.  Information can either be explicitly passed
   in a token, or it can be implicit in the way the Client responds to a
   token challenge.  For example, if a Client fails to complete
   issuance, and consequently fails to redeem a token in response to a
   token challenge, this can reveal information to the Origin that it
   might not otherwise have access to.  
```
- [nms] The example may have poor word choice in that the Client is presumed to know (have access to) all the attributes it discloses in Evidence. Once disclosed, there is no way to un-disclose it. Hence, the wording should say that the Client might disclose attributes to the Origin during the token exchange, that can't be un-disclosed even if the Client aborts the token challenge exchange. 

```
However, an Origin cannot
   necessarily distinguish between a Client that fails to complete
   issuance and one that ignores the token challenge altogether.
```
- [nms] However, the Origin may apply DPA attacks on all tokens such that the attributes disclosed over a period of time may reveal a pattern that is then used to identify / track a particular Client. For example, the time of day, mix of attributes, and number of times a series of failures to complete a token request occurs etc.

### LINKED ISSUE #404: [Secdir review] Shared Origin, Attester, Issuer deployment
```
   This model represents the initial deployment of Privacy Pass, as
   described in [PrivacyPassCloudflare].  In this model, the Attester,
   Issuer, and Origin share the attestation, issuance, and redemption
   contexts.  As a result, attestation mechanisms that can uniquely
   identify a Client, e.g., requiring that Clients authenticate with
   some type of application-layer account, are not appropriate, as they
   could lead to unlinkability violations.
```
- [nms] Can the architecture be more specific about which type of application-layer accounts and which unlinkability violations?  If an application-layer account used a DAA key to attest (sign Evidence) then the unlinkability goal appears to be achieved. Placing a ban on all application accounts without consideration for specific techniques and their privacy preserving properties seems inappropriate.

```
   Origin-Client, Issuer-Client, and Attester-Origin unlinkability
   requires that issuance and redemption events be separated over time,
   such as through the use of tokens that correspond to token challenges
   with an empty redemption context (see Section 3.3), or be separated
   over space, such as through the use of an anonymizing proxy when
   connecting to the Origin.
```
- [nms] The specific attributes that the anonymizing proxy is capable of proxying (hiding) should be enumerated since some attributes are opaque payload elements for a given OSI layer.  A full application layer proxy would, in theory, have the ability to proxy any attribute that is implied by the previous paragraph (yet is banned). 

What causes the Client to trust the proxy to be it's delegate in enforcing the Client's privacy policy vs. being an agent of its adversaries?

### LINKED ISSUE #405: [Secdir review] Joint Attester and Issuer deployment
````
4.2.  Joint Attester and Issuer

   In this model, the Attester and Issuer are operated by the same
   entity that is separate from the Origin.  The Origin trusts the joint
   Attester and Issuer to perform attestation and issue Tokens.  
```
- [nms] Combining these roles seems like it should not be tolerated since the Issuer is an adversary of the Client and the Attester (RATS Verifier) is a trusted ally of the Client. Given the server hosting both roles is under common administrative control, what causes the Client to continue to trust the Attester while not trust the Issuer?

### LINKED ISSUE #406: [Secdir review] Discriminatory Treatment
```
5.1.  Discriminatory Treatment

   Origins can use tokens as a signal for distinguishing between Clients
   that are capable of completing attestation with one Attester trusted
   by the Origin's chosen Issuer, and Clients that are not capable of
   doing the same.  
```
- [nms] Shouldn't the goal of Privacy Pass be that the Client (and it's trusted participants - Attester / RATS Verifier?) are able to opt-out of a privacy revealing exchange before the exchange occurs? Likewise, if the Origin is unable to get a sufficient degree of transparency of the Client, the Origin may opt-out? Given opting-out is a reasonable privacy enforcement strategy, and a threat mitigation strategy, does it make sense to describe this as "discriminatory"?

```
A consequence of this is that Privacy Pass could
   enable discriminatory treatment of Clients based on Attestation
   support.  This could lead to harmful ecosystem effects if left
   unresolved.
```
- [nms]The harmful ecosystem effects should be enumerated.

```
   For example, to disallow discriminatory behavior between Clients with
   and without device attestation support, an Issuer may wish to support
   Attesters that support CAPTCHA-based attestation.  This means that
   the overall attestation value of a Privacy Pass token is bound by the
   difficulty in spoofing or bypassing either one of these attestation
   procedures.
```
- [nms]Wouldn't the value also be tied to the shared attributes that were negotiated? For example, the attributes that are good indicators of Client trustworthiness and are also privacy preserving? Is it within the scope of this I-D to identify some of these attributes - all things considered? If the set of attributes is the empty set, then does it make sense to deploy Privacy
Pass?

### LINKED ISSUE #407: [Secdir review] Centralization
```
5.2.  Centralization

   A consequence of limiting the number of participants (Attesters or
   Issuers) in Privacy Pass deployments for meaningful privacy is that
   it forces concentrated centralization amongst those participants.
   [CENTRALIZATION] discusses several ways in which this might be
   mitigated.  For example, a multi-stakeholder governance model could
   be established to determine what candidate participants are fit to
   operate as participants in a Privacy Pass deployment.  This is
   precisely the system used to control the Web's trust model.

   Alternatively, Privacy Pass deployments might mitigate this problem
   through implementation.  For example, rather than centralize the role
   of attestation in one or few entities, attestation could be a
   distributed function 
```
- [nms]"Attestation" is ambiguous if the reader substitutes RATS "Verifier" with Attester and RATS Attester with Client. Attestation is both the generation (collection) of Evidence and the generation of Attestation Results. The language here suggests the attestation function is distributable, but only in the sense that multiple Verifier instances can be deployed for scalability and the various subroutines of the verifier function can be distributed. The idea that a distributed deployment wouldn't be capable of or have interest in collusion should not be assumed. Rather, mechanisms (such as DAA) should be defined that prevent collusion.

### LINKED ISSUE #408: [Secdir review] Security considerations
```
7.  Security Considerations

   This document describes security and privacy requirements for the
   Privacy Pass redemption and issuance protocols.  It also describes
   deployment models and privacy considerations for using Privacy Pass
   within those models.  Ensuring Client privacy -- separation of
   attestation and redemption contexts 
```
- [nms] Separation of contexts is a reasonable BKM for good privacy architecture, but meaningful implementation should identify enforcement mechanisms that resist collusion attacks.

chris-wood:
Closes #397
Closes #398 
Closes #399
Closes #400
Closes #401 
Closes #402 
Closes #403 
Closes #404 
Closes #405 
Closes #406
Closes #407 
Closes #408


draft-ietf-privacypass-architecture.md
```diff
@@ -126,23 +126,47 @@ the security of all participating entities.
...
 : An entity that attests to properties of Clients for the
-  purposes of token issuance.
+  purposes of token issuance. Using {{?RFC9334}} terminology, Attesters implement
+  the RATS Verifier role.
 
 Attestation procedure:
 : The procedure by which an Attester determines whether or not a Client
   is trusted with a specific set of properties for token issuance.
 
+Anonymization service:
```
tfpauly, line 160: Does this really need to be a definition, since it only used once below?

draft-ietf-privacypass-architecture.md
```diff
@@ -192,8 +216,10 @@ Token Response, the Client computes a token from the token challenge and Token
 Response. This token can be validated by anyone with the per-Issuer key, but
 cannot be linked to the content of the Token Request or Token Response.
 
-6. If the Client has a token, it includes it in a subsequent request to
-   the Origin, as authorization. This token is sent only once.  The Origin
+6. If the Client has a token, it includes it in a subsequent, retried request to
```
tfpauly, line 219: I don't think the request is necessarily a "retried" one — it's a new request with a new header in it. It's in reaction to the challenge, certainly, but I don't know if "retried" is quite right.

draft-ietf-privacypass-architecture.md
tfpauly: If anything, let's move more of this text below where we mention the service.

draft-ietf-privacypass-architecture.md
chris-wood: Hm, true, I suppose it doesn't need to be the exact same request. Will revert.

chris-wood, line 219:
```suggestion
6. If the Client has a token, it includes it in a subsequent request to
```

draft-ietf-privacypass-architecture.md
chris-wood: Done!

tfpauly (approved):

PR #459, opened 2023-08-30
Some editorial suggestions

martinthomson:
These are mostly mechanical in nature.

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -267,7 +267,7 @@ responding to it. Validation requirements are as follows:
 If validation fails, the client MUST NOT process or respond to the
 challenge. Clients MAY have further restrictions and requirements around
 validating when a challenge is considered acceptable or valid. For example,
-clients can choose to ignore challenges that list origin names for which
+clients can choose to ignore challenges that list origin names for which the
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -103,10 +109,12 @@ uses the following terms in more specific ways:
...
-- Token challenge: A requirement for tokens sent from an origin to a client,
-using the "WWW-Authenticate" HTTP header field. This challenge is bound to a
-specific token issuer and issuance protocol, and may be additionally bound to
-a specific context or origin name.
+- Token challenge: A request for tokens sent from an origin to a client, using
+the "WWW-Authenticate" HTTP header field. This challenge identifies a specific
+token issuer and issuance protocol. Token challenges optionally include
+identifier for one or both of: a redemption context (see
+{{context-construction}}), and the origin. These optional identifiers are then
+be bound to the token that is issued.
```
tfpauly, line 117:
```suggestion
token issuer and issuance protocol. Token challenges optionally include
one or both of: a redemption context (see {{context-construction}}), and
a list of associated origins. These optional values are then
be bound to the token that is issued.
```

tfpauly (approved):

PR #461, opened 2023-08-31
Soften language about single-use

### LINKED ISSUE #437: Are tokens really single use?
> Unlike many authentication schemes in which a client will present the same credentials across multiple requests, tokens used with the "PrivateToken" scheme are single-use credentials, and are not reused. Spending the same token value more than once allows the origin to link multiple transactions to the same client. In deployment scenarios where origins send token challenges to request tokens, origins ought to expect at most one request containing a token from the client in reaction to a particular challenge.

I can think of many cases where a client wants to access a series of resources and use this scheme to access each.  I can think of two broad categories:

1. Error recovery.  If the client is unable to retrieve a resource for reasons unrelated to the token (a broken connection, say), they might want to retry the request.
2. Multiple resources within a single transaction or context.  Though it might be possible to only authorize a single request (the root HTML of a web page), HTTP servers often provide access control on a per-resource basis.  Having the token reused in that setting isn't harmful to privacy, particularly if there are other means of correlating requests (like the shared connection that is used to make all those requests).

In other words, I think that a non-reuse requirement could be problematic in the first case and unnecessary in the second.  Of course, loosening this requirement might make replay protections and whatnot more interesting, but that's something that can be managed.

tfpauly:
Closes #437


tfpauly:
@martinthomson please check

martinthomson (approved):

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -119,13 +119,12 @@ clients to present a token from a specific issuer ({{challenge}}). Once a
 client has received a token from that issuer, or already has a valid token
 available, it presents the token to the origin ({{redemption}}).
```
chris-wood, line 120:
```suggestion
available, it presents the token to the origin ({{redemption}}). The process of
presenting a token in response to a challenge is also referred to as "spending" a token.
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -119,13 +119,12 @@ clients to present a token from a specific issuer ({{challenge}}). Once a
...
-credentials across multiple requests, tokens used with the "PrivateToken"
-scheme are single-use credentials, and are not reused. Spending the same token
-value more than once allows the origin to link multiple transactions to the
-same client. In deployment scenarios where origins send token challenges to
-request tokens, origins ought to expect at most one request containing a token
-from the client in reaction to a particular challenge.
+In order to prevent linkability across different transactions, clients
+will often present a particular "PrivateToken" only once. Spending the same
+token value more than once allows the origin to link multiple transactions to
+the same client. In deployment scenarios where origins send token challenges
```
chris-wood, line 125:
```suggestion
will often present a particular "PrivateToken" only once. Origins can link multiple
transactions to the same client if that client spends the same token value more
than once. In deployment scenarios where origins send token challenges
```

chris-wood: Less passive, more active

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -119,13 +119,12 @@ clients to present a token from a specific issuer ({{challenge}}). Once a
...
+to request tokens, origins ought to expect at most one request containing a
+token from the client in reaction to a particular challenge.
```
chris-wood, line 127:
```suggestion
the same client. As such, origins ought to expect at most one unique token
value, carried in one request, for each challenge.
```

This is also true in cases where the client isn't challenged. The suggestion also places emphasis on the uniqueness of the token, rather than of the request, since that's what this whole paragraph is about.

chris-wood (approved):
Approved pending suggestions.

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -117,15 +117,15 @@ Token redemption is performed using HTTP Authentication
...
-from the client in reaction to a particular challenge.
+available, it presents the token to the origin ({{redemption}}). The process of
+presenting a token in response to a challenge is also referred to as "spending" a token.
+
+In order to prevent linkability across different transactions, clients
+will often present a particular "PrivateToken" only once. Origins can link multiple
+transactions to the same client if that client spends the same token value more
+than once. In deployment scenarios where origins send token challenges
+to request tokens, origins ought to expect at most one request containing a
+token from the client in reaction to a particular challenge.
```
tfpauly, line 128:
```suggestion
than once. As such, origins ought to expect at most one unique token
value, carried in one request, for each challenge.
```

draft-ietf-privacypass-auth-scheme.md
tfpauly: Applied in a new commit

PR #464, opened 2023-08-31
Must ignore unknown parameters

### LINKED ISSUE #444: Ignoring unknown parameters
> Clients SHOULD ignore unknown parameters in challenges, except if otherwise specified by issuance protocols.

This should be a "MUST".  Otherwise, we're in a soup of unknown client handling procedures for extensions.

It seems unlikely that extensions will be necessary, but you have not provided any other means of extending, so making this one work is probably wise.

tfpauly:
Closes #444

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -236,8 +236,9 @@ Clients MAY ignore the challenge, e.g., because the token-key is
 invalid or otherwise untrusted.
 
 The header field MAY also include the standard "realm" parameter, if desired.
-Issuance protocols MAY require other parameters. Clients MUST ignore unknown
-parameters in challenges.
+Issuance protocols MAY define other parameters, some of which might be required.
+Clients MUST ignore parameters in challenges that are not defined for the issuance
+protocol corresponding to the token type in the challenge.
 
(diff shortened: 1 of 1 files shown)


chris-wood (changes requested):
I don't agree with this change. We've not yet sorted out how extensions will be negotiated and this might be overly restrictive.

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -236,8 +236,8 @@ Clients MAY ignore the challenge, e.g., because the token-key is
 invalid or otherwise untrusted.
 
 The header field MAY also include the standard "realm" parameter, if desired.
-Issuance protocols MAY require other parameters. Clients SHOULD ignore unknown
-parameters in challenges, except if otherwise specified by issuance protocols.
+Issuance protocols MAY require other parameters. Clients MUST ignore unknown
+parameters in challenges.
```
tfpauly, line 240:
```suggestion
Issuance protocols MAY define other parameters, some of which might be required.
Clients MUST ignore parameters in challenges that are not defined for the issuance
protocol corresponding to the token type in the challenge.
```

chris-wood (approved):

PR #465, opened 2023-08-31
Clarify double-spend state

### LINKED ISSUE #446: Binding requests to connections
> For example, double spend state for unique, per-request redemption contexts only needs to exist within the scope of the request connection or session. 

This is probably not a good idea.  Clients will expect to be able to make a new request over a new connection and will probably not request fresh issuance if a request with a token fails.

(Also, "request connection" is not a concept in HTTP.)

tfpauly:
Closes #446

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -305,9 +305,9 @@ This list is not exhaustive.
 Preventing double spending on tokens requires the origin to keep state
 associated with the redemption context. An empty redemption context is not
 bound to any property of the client request, so state to prevent double spending
-needs to be stored and shared across all servers that can accept request until
+needs to be stored and shared across all origin servers that can accept tokens until
 token-key expiration or rotation. For a non-empty redemption context, the
-double spend state only needs to be stored across the set of servers that will
+double spend state only needs to be stored across the set of origin servers that will
 accept tokens with that redemption context.
 
 Origins that share redemption contexts, i.e., by using the same redemption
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -302,14 +302,13 @@ This list is not exhaustive.
...
 Preventing double spending on tokens requires the origin to keep state
-associated with the redemption context. The size of this state varies based on
-the size of the redemption context. For example, double spend state for unique,
-per-request redemption contexts only needs to exist within the scope of
-the request connection or session. In contrast, double spend state for empty
-redemption contexts must be stored and shared across all requests until
-token-key expiration or rotation.
+associated with the redemption context. An empty redemption context is not
+bound to any property of the client request, so state to prevent double spending
+needs to be stored and shared across all servers that can accept request until
```
chris-wood, line 308:
```suggestion
needs to be stored and shared across all origin servers that can verify and accept requests carrying tokens until
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -302,14 +302,13 @@ This list is not exhaustive.
...
+token-key expiration or rotation. For a non-empty redemption context, the
+double spend state only needs to be stored across the set of servers that will
```
chris-wood, line 310:
```suggestion
double spend state only needs to be stored across the set of origin servers that will
```

tfpauly:
The point here is to remove the reference to the "request connection", and just say "you need to keep the double spend state along with your redemption context", whatever that may be.

chris-wood:
Right, but I don't think that resolves the linked issue.

PR #477, opened 2023-09-07
Clarify server name formats in issuer_name and origin_info

### LINKED ISSUE #440: Issuer identification using an "authority"
> "issuer_name" is an ASCII string that identifies the issuer using the format of the authority portion of a URI as defined in Section 3.2 of [URI].

This is a little surprising to me, which suggests that it might need an explanation.

I might even suggest that this is dangerous, in that it can include `userinfo`, which seems like it might be harmful in this context (though I can't exactly see why).

Would a web origin [RFC6454] be a better choice?

tfpauly:
Closes #440

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -191,8 +191,8 @@ The structure fields are defined as follows:
 - "token_type" is a 2-octet integer, in network byte order, as described
 above.
 
-- "issuer_name" is an ASCII string that identifies the issuer using the format of the authority portion of a URI
-as defined in {{Section 3.2 of !URI=RFC3986}}. This name identifies the issuer that is allowed to
+- "issuer_name" is an ASCII string that identifies the issuer using the format
+of a server name described below. This name identifies the issuer that is allowed to
 issue tokens that can be redeemed by this origin. The field that stores this string in the challenge
 is prefixed with a 2-octet integer indicating the length, in network byte order.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -205,12 +205,19 @@ practice. Challenges with redemption_context values of invalid lengths MUST be i
...
-commas "," without any whitespace. If this field is not empty, the Origin MUST
-include its own name as one of the names in the list.
+origin name uses the format of a server name described below. The string is
+prefixed with a 2-octet integer indicating the length, in network byte order.
+If empty, any non-origin-specific token can be redeemed. If the string contains
+multiple origin names, they are delimited with commas "," without any whitespace.
+If this field is not empty, the Origin MUST include its own name as one of the
+names in the list.
+
+The server names used in the "issuer_name" and "origin_info" fields are ASCII strings
```
chris-wood, line 215: I think this would read better if we put this description in the terminology section and then referenced it in these two spots.

chris-wood (approved):

draft-ietf-privacypass-auth-scheme.md
tfpauly: @chris-wood I tried that, but it felt awkward in terminology.

I've now updated the sections for challenges and redemption to include more subsections, and gave this its own subsection to refer to. I think this should make it cleaner to read. Thoughts?

tfpauly: See https://ietf-wg-privacypass.github.io/base-drafts/tfpauly-patch-34/draft-ietf-privacypass-auth-scheme.html

chris-wood: Love it!

PR #478, opened 2023-09-12
Structure the security considerations

### LINKED ISSUE #453: Structure the security considerations
There are several different concepts here.  Add section headings for the following:

* Replay attacks (integrity)
* Token exhaustion attacks (DoS)
* Timing correlation attacks (privacy)
* Cross-context recognition

chris-wood:
Closes #453

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -553,7 +553,7 @@ scheme described in this document.
 
 ## Randomness Requirements
 
-Unless otherwise stated, all random values in the challenge and token MUST be
+All random values in the challenge and token MUST be
 generated using a cryptographically secure source of randomness ({{?RFC4086}}).
 
 ## Replay Attacks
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -548,6 +548,16 @@ redeem tokens is maintained and exercised consistently.
 
 # Security Considerations {#sec-considerations}
 
+This section contains security considerations for the PrivateToken authentication
+scheme described in this document.
+
+## Randomness Requirements
+
+Unless otherwise stated, all random values in the challenge and token MUST be
```
tfpauly, line 556: Are there places where we state otherwise? Or can we just say "all..."

tfpauly (approved):

draft-ietf-privacypass-auth-scheme.md
chris-wood: Sort of... when we talk about how to construct the redemption context, we say it can be random, but can also be something else. We are probably fine without the precondition though, so I'll just drop it.

chris-wood, line 556:
```suggestion
All random values in the challenge and token MUST be
```

PR #479, opened 2023-09-12
Add overview content to token challenge section

### LINKED ISSUE #438: The token challenge could use an overview
Section 2.1 describes the use of "PrivateToken" challenges in WWW-Authenticate (site note: and proxy-authenticate?), which is fine.

It then launches into a description of what a TokenChallenge message includes, failing to note some important stuff up front:

1. the field value is a series of key-value pairs, with values being base64-encoded binary data
2. there are `challenge` and `token-key` keys that are both required
3. the challenge can be repeated

The necessary information is all there, but it could really use a map.

chris-wood:
Closes #438

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -140,23 +140,31 @@ in more detail.
 ## Token Challenge {#challenge}
 
 Origins send a token challenge to clients in an "WWW-Authenticate" header field
-with the "PrivateToken" scheme. This challenge includes a TokenChallenge
-message, along with information about what keys to use when requesting a token
-from the issuer.
+with the "PrivateToken" scheme. This field is a series of key-value pairs, where
+values are base64url-encoded binary data. Each field has two mandatory parameters,
+one containing a token challenge and another containing the token-key used for
+producing (and verifying) a corresponding token. The challenge parameter can be
+repeated in the header field.
 
-Origins that support this authentication scheme need to handle the following
-tasks:
+Origins that support the "PrivateToken" authentication scheme need to handle
+the following tasks in constructing the WWW-Authenticate header field:
 
 1. Select which issuer to use, and configure the issuer name and token-key to
-include in WWW-Authenticate challenges.
+include in WWW-Authenticate token challenges. The issuer name is included in
+the token challenge, and the issuer token-key is used to populate the
+WWW-Authenticate header parameter.
 
 1. Determine a redemption context construction to include in the
-TokenChallenge, as discussed in {{context-construction}}.
+token challenge, as discussed in {{context-construction}}.
 
-1. Select the origin information to include in the TokenChallenge. This can
+1. Select the origin information to include in the token challenge. This can
 be empty to allow fully cross-origin tokens, a single origin name that
-matches the origin itself, or a list of origin names containing the origin
-itself.
+matches the origin itself for per-origin tokens, or a list of origin names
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -140,23 +140,31 @@ in more detail.
 ## Token Challenge {#challenge}
 
 Origins send a token challenge to clients in an "WWW-Authenticate" header field
-with the "PrivateToken" scheme. This challenge includes a TokenChallenge
-message, along with information about what keys to use when requesting a token
-from the issuer.
+with the "PrivateToken" scheme. This field is a series of key-value pairs, where
+values are base64url-encoded binary data. Each field has two mandatory parameters,
```
tfpauly, line 144: Not all parameters to the auth scheme will necessarily be base64... I wonder if we shouldn't go into the key-value bit, since that's normal to an auth scheme.

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -140,23 +140,31 @@ in more detail.
 ## Token Challenge {#challenge}
 
 Origins send a token challenge to clients in an "WWW-Authenticate" header field
```
tfpauly, line 142: I wonder if we should mention "(or "Proxy-Authenticate")

chris-wood: That seems like an increase in scope, so if we do we'd probably want to run it past the list.

draft-ietf-privacypass-auth-scheme.md
chris-wood, line 144:
```suggestion
with the "PrivateToken" scheme. This field is a series parameters, each a key-value pair. There are two mandatory parameters:
```

draft-ietf-privacypass-auth-scheme.md
tfpauly: Yeah we don't necessarily need to mention that — it was just alluded to in one of martin's comments

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -140,23 +140,30 @@ in more detail.
...
-with the "PrivateToken" scheme. This challenge includes a TokenChallenge
-message, along with information about what keys to use when requesting a token
-from the issuer.
+with the "PrivateToken" scheme. This field is a series parameters, each a key-value pair. There are two mandatory parameters:
+one containing a token challenge and another containing the token-key used for
```
tfpauly, line 144:
```suggestion
with the "PrivateToken" scheme. This authentication scheme has two mandatory parameters:
one containing a token challenge and another containing the token-key used for
```

tfpauly: Since this is just part of how auth schemes work...

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -140,23 +140,30 @@ in more detail.
...
+producing (and verifying) a corresponding token. The challenge parameter can be
+repeated in the header field.
```
tfpauly, line 146: The challenge parameter isn't repeated, the entire authentication challenge with all parameters is repeated. Again, this is just how auth schemes work.

```suggestion
producing (and verifying) a corresponding token.
```

chris-wood: Right, right.

tfpauly (approved):

PR #480, opened 2023-09-12
Make reference to the PrivateToken object

armfazh:
The `Authorization:` header contains a `Token` object, as well as the `WWW-Authenticate:` header contains a `PrivateToken` object.

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -78,7 +78,7 @@ interaction between client and origin is shown below.
 | Origin |                              | Client |
 +---+----+                              +---+----+
     |                                       |
-    +-- WWW-Authenticate: TokenChallenge -->|
+    +-- WWW-Authenticate: PrivateToken ---->|
     |                                       |
     |                            (Run issuance protocol)
     |                                       |


chris-wood (commented):
Hmm, this isn't quite right. The challenge header contains a TokenChallenge.

tfpauly (changes requested):
"PrivateToken" is the auth scheme. The contents of the challenge are a "TokenChallenge" struct, and the redemption contains a "Token" struct.

armfazh:
> Hmm, this isn't quite right. The challenge header contains a TokenChallenge.

But it is wrapped in a `PrivateToken` object, isn't it?

> "PrivateToken" is the auth scheme.

Correct, this is why I'm proposing changes to the auth-scheme document.

tfpauly:
@armfazh there is no such thing as a "PrivateToken" object defined here. It's just an auth scheme name, not an object or struct.

armfazh:
Ok, I think I understand it better, and just to clarify.

The authentication scheme is called `PrivateToken` and during the initial request looks like:
```
WWW-Authenticate:
  PrivateToken challenge="abc...", token-key="123..."
                          ^ 
                          | this is the encoded TokenChallenge struct
```

And during redemption, client sends:
```
Authorization: PrivateToken token="abc..."
                                   ^
                                   | this is the Token struct 
```

So the diagram is not exactly showing the auth scheme, but instead the messages that are send through it. So, it's a simplified version.

tfpauly:
That's right!

PR #481, opened 2023-09-12
Clarify replay and reflect

### LINKED ISSUE #454: Replay attacks via 0-RTT
There is no inherent problem with a replay of a token.  A problem only exists if replay occurs and the request that it authorizes has some sort of unwanted side effect.  That's not a problem with this document, so I'm not sure that you need to cite that document.

chris-wood:
Closes #454

### DIFF
--- a/draft-ietf-privacypass-auth-scheme.md
+++ b/draft-ietf-privacypass-auth-scheme.md
@@ -450,9 +450,9 @@ be bound to a specific TLS session with a client, but origins can also accept
 tokens for valid challenges in new sessions. Origins SHOULD implement some form
 of double-spend prevention that prevents a token with the same nonce from being
 redeemed twice. Double-spend prevention ensures that clients cannot replay tokens
-for previous challenges. For context-bound tokens, this double-spend prevention
-can require no state or minimal state, since the context can be used to verify
-token uniqueness.
+for previous challenges. See {{replay-attacks}} for more information about replay
+attacks. For context-bound tokens, this double-spend prevention can require no state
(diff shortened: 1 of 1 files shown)


draft-ietf-privacypass-auth-scheme.md
```diff
@@ -587,33 +587,39 @@ generated using a cryptographically secure source of randomness ({{?RFC4086}}).
 
 ## Replay Attacks
```
tfpauly, line 588:
```suggestion
## Replay Attacks {#replay-attacks}
```

draft-ietf-privacypass-auth-scheme.md
```diff
@@ -587,33 +587,39 @@ generated using a cryptographically secure source of randomness ({{?RFC4086}}).
...
 
+Applications SHOULD constrain tokens to a single origin unless the use case can
+accommodate such replay attacks. Replaying tokens is not necessarily a security
```
tfpauly, line 591:
```suggestion
accommodate replay attacks. Replaying tokens is not necessarily a security
```

Since we don't list the attacks before, they're not "such"

tfpauly (approved):

draft-ietf-privacypass-auth-scheme.md
chris-wood: Is this for consistency? (This isn't needed for the build -- the anchor is automagically included if not explicitly written.)

tfpauly: Ah okay, then you can skip it

PR #484, opened 2023-09-12
Clarify some token properties

### LINKED ISSUE #472: Clarifications in Token Type Registry definitions
(1) p 15, sec 5.2.  Token Type Registry

  *  Private Metadata: A Y/N value indicating if the output tokens can
     contain private metadata.

This is the first time that some of these fields (e.g., Publicly Verifiable,
Public/Private Metadata) have been introduced.  Does the document need any
additional prose to describe what they and how they are used?  The current text
feels somewhat terse as a description in a standard track document.

(2) p 17, sec 5.2.  Token Type Registry

  *  Nid: N/A

Shoudln't Nk and Nid default to 0 rather than 'N/A'?  This comment also applies
to the text above the greased values, or otherwise (at a stretch) it could
arguably be interpreted as putting is randomly sized Nk and Nid fields
containing random data.

chris-wood:
Closes #472


tfpauly (approved):
I assume we're happy leaving `Nid: N/A` for the greased values, yes?

chris-wood:
> I assume we're happy leaving `Nid: N/A` for the greased values, yes?

Yeah, since they're sort of meaningless. I'd also be happy with 0. Honestly, whatever IANA folks prefer!

PR #492, opened 2023-09-21
Address Murray's IESG review of the protocol document

chris-wood:


draft-ietf-privacypass-protocol.md
```diff
@@ -179,7 +179,8 @@ Beyond staging keys with the "not-before" value, Issuers MAY advertise multiple
 Issuers indicate preference for which token key to use based on the order of
 keys in the list, with preference given to keys earlier in the list. Clients
 SHOULD use the first key in the "token-keys" list that either does not have a
-"not-before" value or has a "not-before" value in the past. Origins can attempt
+"not-before" value or has a "not-before" value in the past, as this key is most
+likely to be valid in the given time window. Origins can attempt
```
tfpauly, line 183:
```suggestion
"not-before" value or has a "not-before" value in the past, since the first such key is the
most likely to be valid in the given time window. Origins can attempt
```

tfpauly: Is this clearer?

tfpauly (approved):

draft-ietf-privacypass-protocol.md
chris-wood: Yep!

PR #505, opened 2023-12-19
Fix typos in protocol spec

### LINKED ISSUE #504: Typos in protocol document
From a reviewer of the protocol doc:

> https://datatracker.ietf.org/doc/draft-ietf-privacypass-protocol/
> Section 6.3 has probably two errors:
> First one:
> authenticator =
>      Finalize(pkI, nonce, blind_sig, blind_inv)
> I think that instead of nonce we need to pass token_input to the Finalize:
>    token_input = concat(0x0002, // Token type field is 2 bytes long
>                         nonce,
>                         challenge_digest,
>                         token_key_id)
> Otherwise the Finalize function fails to verify the signature.
> Second:
> The Token.nonce value is that which was sampled in Section 5.1
> I think this one should be:
> ... sampled in Section 6.1

chris-wood:
Closes #504

### DIFF
--- a/draft-ietf-privacypass-protocol.md
+++ b/draft-ietf-privacypass-protocol.md
@@ -660,7 +660,7 @@ content as follows:
 
 ~~~
 authenticator =
-  Finalize(pkI, nonce, blind_sig, blind_inv)
+  Finalize(pkI, PrepareIdentity(token_input), blind_sig, blind_inv)
 ~~~
 
 The Finalize function is defined in {{Section 4.4 of BLINDRSA}}. If this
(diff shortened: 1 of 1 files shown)


tfpauly (approved):

paulwouters:
Note that I wouldn’t call this change a “typo”. It’s a substantive change of the nonce parameter.Sent using a virtual keyboard on a phoneOn Dec 19, 2023, at 11:47, Tommy Pauly ***@***.***> wrote:﻿
Merged #505 into main.

